PR #14, opened 2022-06-20
Please verify and update QoD def to v. 1.2.1

T-sm:
Please verify and update QoD definitions to v. 1.2.1


T-sm:
Removed apiKey entry in order to be compliant with Camara API Authentication and Authorization document.

shilpa-padgaonkar:
Obsolete

Issue #42, opened 2022-09-21
QoD API specification version 0.8.0 


akoshunyadi:
Contribution of the QoD-API spec v0.8.0

It contains following changes based on v0.1.0:

1.	Two APIs (latency & throughput) combined to one, 0.8.0 replaces 0.1.0
2.	ueAddr (ipv4) changed to ueId (ipv4, ipv6, msisdn, externalid)
3.	asAddr (ipv4) changed to asId (ipv4, ipv6)
4.	Portspec with schema
5.	Protocol (in & out) removed 
6.	QoS profiles renamed
7.	Basepath changed
8.	Some formal changes/fixes

The existing yaml-files for latency and throughput should be removed.
The profile mapping file should be adjusted.

SyeddR:
The new spec looks good to me. Couple of ques

1. Are we expecting both ipv4 and ipv6 information to be provided in API? If only one type is required shall we use "anyof" in the spec? If both IPs are provided, is the expectation to create QoS for both ipv4 and ipv6 session?
2. There is a TFT limit device (UE) can support number of flows. If we are allowing port ranges and ports that would easily reach TFT limit. Are we going to enforce restriction in the API or is it left to operator's implementation

eric-murray:
Hi @SyeddR 

The idea is that the API consumer provides what information they know about the UE and AS sufficient to set-up the required session. If they wanted the enhanced QoS to apply to both IPv4 and IPv6 connectivity, then they would need to provide both addresses, at least for the AS. For the UE, any one of the 4 specified identifiers can identify the UE itself, though I agree this may lead to ambiguity as to intentions.

For example, if the API consumer provides both IPv4 and IPv6 addresses for the AS, but only an IPv6 address for the UE, does that mean they want enhanced QoS for both IPv4 and IPv6 flows, and only provided the UE IPv6 address to allow the UE to be identified, or do they only want the enhanced QoS for the IPv6 flow (in which case, why provide the AS Ipv4 address)? I think a lot of details about behaviour will come down to individual operator implementations of the API, and we cannot expect consistent behaviour across different implementations for all possible parameter values.

[...]

Issue #7, opened 2021-12-15
Add new endpoint in QoD API to request available qos_profiles


shilpa-padgaonkar:

RandyLevensalor:
> It would be great to have an API to list the available QOS profiles which are available for a specific user and / or device. The list of QOS profiles would include the attributes for each profile.

The query extensions are in issue #147

Go ahead and close this issue

hdamker:
As agreed in call ... remaining functionality in #147

PR #43, opened 2022-09-21
Contribution of the QoD-API spec v0.8.0

akoshunyadi:
PR for the issue #42 

The QoD-API spec 0.8.0 contains following changes based on QoD API-spec v0.1.0:

- Two APIs (latency & throughput) combined to one, 0.8.0 replaces 0.1.0 
- ueAddr (ipv4) changed to ueId (ipv4, ipv6, msisdn, externalid) 
- asAddr (ipv4) changed to asId (ipv4, ipv6)
- Portspec contains now a schema
- Protocol (in & out) removed
- QoS profiles renamed
- Basepath changed
- Some formal changes/fixes

The separated API specs (latency and throughput) have been removed.

### DIFF
--- a/documentation/API_documentation/QoD_Stable_Bandwidth_API.md
+++ b/documentation/API_documentation/QoD_Stable_Bandwidth_API.md
@@ -31,7 +31,7 @@ Define throughput requirements of the application and identify QoS profile class
 Describes the precise flow the developer wants to prioritize and have stable bandwidth for. This flow is described using source and destination IP addresses, ports and protocols with flow direction.
 
 **Duration**
-Define the number of seconds for which the QoD session should be created. This parameter is optional and if not specified, the session is either deleted on user request or if default expiration limit has been reached (24 hours in reference implmentation).
+Define the number of seconds for which the QoD session should be created. This parameter is optional and if not specified, the session is either deleted on user request or if default expiration limit has been reached (24 hours in reference implementation).
 
 **Notification URL and token**
 Developers have a chance to specify callback URL on which notifications (eg. session termination) regarding the session can be received from the service provider. This is also an optional parameter.


eric-murray (approved):
I'm happy with the proposed changes

patrice-conil:
If you remove the references to protocolIn and protocolOut there is no longer any reference to the protocol. What protocol are you taking into account in this case ?

eric-murray:
@patrice-conil 
> If you remove the references to protocolIn and protocolOut there is no longer any reference to the protocol. What protocol are you taking into account in this case ?

All protocols are allowed in both directions

maheshc01:
these are good changes. Completely agree with the need for merging QoS and QoD API spec.
Thank you for updates.

Few inputs listed below

1. parameter name update suggestion to make it more developer friendly 
change "asId" to "applicationServerEndpoint"

2. Make uePorts part of ue ID
Make asPorts part of applicationServerEndpoint(new name for asId)
Sample:
"ueId": {
    "externalId": "123456789@domain.com",
    "msisdn": "123456789",
    "ipv4addr": "192.168.0.1/24",
    "ipv6addr": "2001:db8:85a3:8d3:1319:8a2e:370:7344",
	"port": {
    "ranges": [
      {
        "from": 5010,
        "to": 5020
      }
    ],
    "ports": [
      5060,
      5070
    ]
  }
  },
  "asId": {
    "ipv4addr": "192.168.0.1/24",
    "ipv6addr": "2001:db8:85a3:8d3:1319:8a2e:370:7344",
	"uePorts": {
    "ranges": [
      {
        "from": 5010,
        "to": 5020
      }
    ],
    "ports": [
      5060,
      5070
    ]
  }
  },
makes it clear which ports are associated to what and nice way to group related data.

3. Introduce fallbackQos incases where preferred qos setting fails, rather than default, we can offer an alternate option as supported by SCEF.

4. update response message for 201 response code of POST /session to just provide SessionId and Status instead of passing all the input data.

4. In the GET /sessions response, lets add a status field to let the caller of the API know of the current status of the session. 
Status could be one of the following:
- pending_active
- active
- failed
- deleted
- pending_delete
- terminated

5. The API spec does not support PUT API. Any specific reason for not having it?
What happens if some parameter as passed in create session changes like, application server ip or something else?

6. Why not send notification for all status change events rather than just SESSION_TERMINATED?

hdamker:
@emil-cheung within call on Friday Joachim mentioned that you have done a review with your colleagues. Would be great if you add your comments and/or approval.

emil-cheung (changes requested):
1. Currently "ueId" provides 4 parameters and sets minProperties as 1. Per we understand, it should be oneOf, right? Otherwise, for example, if msisdn and ipv4addr are provided at the same time and they are mismatched, how Camara layer is supposed to handle?
2. The "qos" parameter is not future proof. Only the values listed in the enumeration are valide.
Suggest changing to a more future compliant definition.

QosProfile:
    anyOf:
     - type: string
      enum:
        - QOS_E
        - QOS_S
        - QOS_M
        - QOS_L
      description: |
        * `QOS_E` - Qualifier for enhanced communication profile
        * `QOS_S` - Qualifier for the requested QoS profile _S_
        * `QOS_M` - Qualifier for the requested QoS profile _M_
        * `QOS_L` - Qualifier for the requested QoS profile _L_
    - type: string
    -  description: >
          This string provides forward-compatibility with future
          extensions to the enumeration but is not used to encode
          content defined in the present version of this API.

akoshunyadi:
@emil-cheung 
1. the idea was that you pass all IDs you have and the service can take, what it needs. It means different implementations can need different identifier, but the API call can remain unchanged
2. I think using a free string without a discovery capability is just a "guessing game"

eric-murray:
@emil-cheung 
I think we may need a discussion on how to handle dual-stack UEs. Issues I see are:
- The API consumer may want both IPv4 and IPv6 flows prioritised, or only one
- If the API consumer only wants the IPv4 flow prioritised, but all they know is the private address, then they will anyway need to provide some additional parameter to identify the UE (such as MSISDN)

Restricting the UE identifier to a single parameter would rule out these scenarios. If an API consumer supplies multiple inconsistent parameters, then clearly the API should return an error - it is not for the API provider to guess what the API consumer really wanted. If the API consumer does not want to risk such rejection, then the API definition allows them to specify only a single parameter.

hdamker:
@emil-cheung

1. I guess this point is answered by @eric-murray and @akoshunyadi ... multiple identifier are allowed and make sense in certain scenarios, the documentation should explain this. If the the API provider detects inconsistencies it will reject the request.
2. Forward compability for upcoming changes would be great but at the current point we have several issues open which will anyway bringn breaking changes, e.g. the discovery function (#7). Therefore I wouldn't introduce a wildcard now into the qos parameter but later a solution which can be used in conjunction with the discovery function.

@emil-cheung Are your comments resolved by the discussion?
@akoshunyadi Can you please your branch with HEAD of main so that it is possible to merge without conflict?

akoshunyadi:
Hi,

Sorry, because of some technical issues I had to create new PR for the 0.8.0
https://github.com/camaraproject/QualityOnDemand/pull/54

KR

Issue #61, opened 2022-10-25
QoSProfile_Mapping_Table.md should be improved


hdamker:
As noted in [MOM-2022-10-07.md](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/MeetingMinutes/MOM-2022-10-07.md) the [QoSProfile_Mapping_Table.md](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/API_documentation/QoSProfile_Mapping_Table.md) should be improved:
> - The note should emphasize that it is just an example how the QoS Profiles can be mapped and not binding for operators.
> - There is a notion that guranteed bit rate (GBR) can be offered only in limited scenarios (e.g. campus). The table is currently providing only examples for relative prioritization.
> - Proposals for changed wording (e.g. "up to xx Mbit" replaced with more use case oriented wording) are welcome.

ToshiWakayama-KDDI:
Sorry for being late, but I have some comments to this and would like to discuss during today's call. Thanks.

Issue #32, opened 2022-08-02
Resolving the confusion of “protocolIn”/“protocolOut” with a better parameter design


emil-cheung:
[Relevant to Issue #13](https://github.com/camaraproject/QualityOnDemand/issues/13)

The current “protocolIn”/“protocolOut” parameters are confusing. Considering the IP protocols:

- TCP is a bi-directional protocol on the ​same socket pairs
- UDP can be used uni-directional or bi-directional on ​the same socket pairs

It is a bit confusing to specify different protocols in the same IP flow for different directions.

eric-murray:
@emil-cheung 
I think "confusing" is the wrong term to use here. Rather, allowing the API consumer to separately specify protocolIn and protocolOut opens up the possibility that they could request an illogical configuration, such as protocolIn = TCP and protocolOut = UDP. To me, that is an error and the API should respond as such, but it would be better if this possibility was removed.

My preference would be to adopt the RFC 6733 convention of separately specifying Protocol and Direction (with the additional option of "both" as well as "in" and "out") for an IPFilterRule. The **BIG** issue with the current definitions is that there is no "NONE" option for the protocol. So if you want to watch 4K video and request THROUGHPUT_L on the downlink for this, the current API definition means you will also get THROUGHPUT_L on the uplink, which is an unnecessary expense. Specifying "Protocol" and "Direction" separately would allow the high throughput to be requested only for the downlink.

The other issue I have with both the current definition (protocolIn and protocolOut) and separately specifying Direction is that "in" and "out" might be considered ambiguous. I've spent my whole career thinking in terms of "uplink" and "downlink", whilst in North America they think in terms of "forward" and "reverse", but again these definitions might not be obvious to a developer. As we explicitly identify the "UE" and "AS", I'd suggest the Direction options should be "UEtoAS", "AStoUE" or "BOTH".

hdamker:
I propose that we use this issue to discuss the paramter design for "in" and "out" and closing the issue #35 which is from my perspective a duplicate.

patrice-conil:
@eric-murray I totally agree with your proposal to adopt RFC 6733 convention.

emil-cheung:
@eric-murray 
The intention to name 'remoteAddr' is not due to UE calling the API.b
It is to generalize 'UE-to-AS' communication as well as 'UE-to-UE' (defined by 3GPP and may applicable to dedicated networks or private networks).

emil-cheung:
[...]

PR #62, opened 2022-11-01
Improvement proposal for QoSProfile_Mapping_Table.md

tlohmar:
The meaning of the different QoS Profile Labels are pretty unclear for a developer, who has no insights into the 5G Network realization (e.g. the meaning of 5QI ==6)

The table is extended with two examples, describing the meaning of the QOS Profile Labels and also constrains, in which cases the Network Service cannot be provided.


documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -1,11 +1,25 @@
 # QoS Profiles Mapping Table (REFERENCE DRAFT)
 
-| Profiles | User Info | 5Qi mapping |
+Here are two possible examples of mapping tables from QoS Profiles to provided connectivity characteristics. The second example describes a case, where the network is specially prepared in well define geographical areas for supporting Quality on Demand needs.
+
+Example 1: Considering wide-area network usage.
+| QOS Profile labels | Network Service Description | Network Service Constraints |
+|----------| --------- | ----------- |
+| QOS\_E   | Latency stays stable under congestion (throughput upto 2Mbps). The application bitrate should not exceed 2Mbps. | The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
```
hdamker, line 8:
```suggestion
| QOS\_E   | Latency stays stable under congestion (throughput upto 500kbps). The application bitrate should not exceed 500kbps. | The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
```
The intention of this QoS profile is to support use case which need a reliable connectivity even under congestion but with a low bit rate.

The sentence "The access to the network service is restricted to a maximal number of simultaneous session." fits better to the QoS profiles with higher bandwidth (e.g. QoS_L).

"The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement." ... can be read differently (geographical region within a network? A region containing several networks?). In addition suggesting to delete pointers to SLAs.

documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -1,11 +1,25 @@
...
+The QOS profile labels in the table can be used within Camara for validating the QoD APIs.
+
+Example 2: Considering confined area network usage.
+| QOS Profile labels | Network Service Description | Network Service Constraints |
 |----------| --------- | ----------- |
-| QOS\_E   | Latency stays stable under congestion (throughput upto 2Mbps) | 7 |
-| QOS\_L   | DL upto 100Mbps (unlimited?) | 6 |
-| QOS\_M   | DL upto 30Mbps | 6 |
-| QOS\_S   | DL upto 10Mbps | 6 |
+| QOS\_E   | Latency stays stable under congestion (throughput upto 2Mbps). The application bitrate should not exceed 2Mbps. | The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
```
hdamker, line 19:
```suggestion
| QOS\_E   | Latency stays stable under congestion (throughput upto 500kbps). The application bitrate should not exceed 500kbps. | The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
```
Plus same comments as above.

documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -1,11 +1,25 @@
 # QoS Profiles Mapping Table (REFERENCE DRAFT)
 
-| Profiles | User Info | 5Qi mapping |
+Here are two possible examples of mapping tables from QoS Profiles to provided connectivity characteristics. The second example describes a case, where the network is specially prepared in well define geographical areas for supporting Quality on Demand needs.
+
+Example 1: Considering wide-area network usage.
+| QOS Profile labels | Network Service Description | Network Service Constraints |
+|----------| --------- | ----------- |
+| QOS\_E   | Latency stays stable under congestion (throughput upto 2Mbps). The application bitrate should not exceed 2Mbps. | The service is only provided, when the UE is located within the geographical area, as agreed in the Service Level Agreement.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
+| QOS\_L   | DL upto 100Mbps (unlimited?). The 5G System throughput is prioritized up to 100Mbps. At high load, the throughput is capped at 100Mbps and can be degraded to lower effective throughput (without a minimum level).  | The service is only provided, when the UE is located within the home network. |
```
hdamker, line 9:
```suggestion
| QOS\_L   | DL upto 100Mbps. The 5G System throughput is prioritized up to 100Mbps. At high load, the throughput is capped at 100Mbps and can be degraded to lower effective throughput (without a minimum level).  | The service is only provided, when the UE is located within the home network. |
```
Suggestion to delete "(unlimited?)"

"The service is only provided, when the UE is located within the home network." ... that is an operator specific decision, e.g. can be offered across multiple subsidiaries and/or also in partner networks.

hdamker (changes requested):
Thanks for the proposal. Full agreement with the general direction to describe the QoS profiles from developer perspective and to include also potential restrictions.

A more explicit note that these are only examples and the actual service offerings by network operators will differ would be good.

DL might be not understandable for non-telco experts (see discussion in #32). And a statement about both traffic directions would be helpful (e.g. QoS_S/M/L only for applications server to user device, while QoS_E is supporting both directions).



See minor suggestions in the comments.

documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -1,11 +1,25 @@
...
+| QOS\_M   | The 5G System throughput is assured for at least 5Mbps and up to 15Mbps. Higher bitrates are possible, but not assured | The service is only provided, when the UE is located in the geographical area as agreed as part of the Service Level Agreement. When the UE is located outside of the geographical area, the 5G System rejects the requested QOS profile.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
+| QOS\_S   | The 5G System throughput is assured for at least 500kbps and up to 1Mbps. Higher bitrates are possible, but not assured. | The service is only provided, when the UE is located in the geographical area as agreed as part of the Service Level Agreement. When the UE is located outside of the geographical area, the 5G System rejects the requested QOS profile.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
```
eric-murray, line 22: What do you mean by "assured" here? What is expected to happen if the minimum bitrate cannot be met in parts of the agreed geographical area because of the radio conditions in which the end user device finds itself (for example, if the end user requests a service for central London, but then gets on an underground train)? Given that mobile operators cannot break the laws of physics, there will be circumstances under which the minimum bitrate cannot be assured in any location, and if the API caller accepts that it cannot be completely assured, then it just becomes a "best effort" service.

Also, how is the agreed geographical area to be specified (this is not part of the current API definition), and how is the mobile operator expected to track the current location of the end user device for the purposes of determining if the assurance has been met or not? If this is based on a group of cells (base-stations), then the end user on the underground train will say "I was in central London and I didn't get my 20 Mb/s", and we'll say "You weren't connected to one of the agreed cells". Technically, we would be correct, but the end user would perceive that the service was not delivered as advertised.

documentation/API_documentation/QoSProfile_Mapping_Table.md
tlohmar: It is certainly ok to lower the bitrates. Still, some admission control might still be needed, even when lowering the bitrates.  Thus, adding this sentence "The access to the network service is restricted to a maximal number of simultaneous session."  to QoS_L is ok. However, not sure about removing it from QOS_E.

The definition of the geographical region is intentionally a bit blurry. One geographical region can be "the home network" or even larger. When deleting the SLA, does this mean that there is no SLA or such geographical area definition is not mentioned in the SLA?

documentation/API_documentation/QoSProfile_Mapping_Table.md
tlohmar: ok, that it a left over from the original table

hdamker:
Call on 2022-11-04: comments to be addressed, PR will address #61

eric-murray (approved):
Still a few typos, but meaning is clear

documentation/API_documentation/QoSProfile_Mapping_Table.md
tlohmar: The pull request should include these comments now. Please check.

documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -1,25 +1,25 @@
...
-| QOS\_S   | DL upto 10Mbps. The 5G System throughput is dprioritized up to 10Mbps. At high load, the throughput is capped at 10Mbps and can be degraded to lower effective throughput (without a minimum level). | The service is only provided, when the UE is located within the home network. |
+| QOS\_E   | Latency stays stable under congestion (throughput upto 500kbps). The application bitrate should not exceed 500kbps. | The service is only provided, when the UE is located within the geographical area, according to the agreement with the CSP.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
+| QOS\_L   | The 5G System throughput is prioritized up to 20Mbps. At high load, the throughput is capped at 20Mbps and can be degraded to lower effective throughput (without a minimum level).  | The service is only provided, when the UE is located within the home network.<br>The access to the network service is restricted to a maximal number of simultaneous session.|
```
hdamker, line 11: "The access to the network service is restricted to a maximal number of simultaneous session." is currently added only to QOS_E and QOS_L ... but should be consequently also added to QOS_M and QOS_E

hdamker (approved):
Content ok, just one comment for consistency.

documentation/API_documentation/QoSProfile_Mapping_Table.md
tlohmar: ok, I added this. thx,

hdamker (approved):

eric-murray (approved):

ToshiWakayama-KDDI:
Sorry for being late, but I have some comments to this and would like to discuss during today's call.  Thanks.

jlurien:
In TEF we are also discussing internally this issue and would like to comment on it. Sorry, not aware of a deadline.

ToshiWakayama-KDDI:
Hi,
I raised some issues on Mapping table at the last QoD API meeting on 18th Nov, and it is supposed to be discussed by email discussion.  I would like Herbert to start email discussion on the Mapping table, or if not, I would propose to re-open this PR and discuss it.
Thanks.

Issue #38, opened 2022-08-12
Add SUCCESSFUL_RESOURCES_ALLOCATION to QosChanged


patrice-conil:
Add the possibility to send the SUCCESSFUL_RESOURCES_ALLOCATION event to notify that everything went as planned as discussed in #33

hdamker:
Closed with merge of #67

Issue #33, opened 2022-08-04
Small API modifications to reduce resources consumption


patrice-conil:
Hello everyone,
When I implement a facade pattern in front of an API, I expect to be able to do it in a stateless way...just using mapping techniques
For our QOD API in front of SCEF this is not possible due to small design decisions.

-SessionInfo => AssessionWithQoSSubscription
OK

- AssessionWithQoSSubscription => InfoSession
We can't set startedAt and duration, but we can set expiresAt
Since the requestor knows when they requested the QOD and for how long, we can probably remove those fields and just keep the expiresAt
The SessionInfo id is defined as uuid... which cannot match the id part in the self field of ofAsSessionWithQoSSubscription.
If we define it as a string, we can use the underlying identifier from SCEF.

-NotificationData => QosChanged KO
Here the problem comes from the qod parameter that cannot be retrieved from NotificationData. 
But the requestor must be able to match is initial request with the request id ... so I think we can remove it from QosChanged.

With these small modifications that don't impact functionalities IMHO we can implement stateless service that will be more efficient.
What do you think about?

Patrice

eric-murray:
Hi @patrice-conil 

I agree stateless design is an admirable goal. Can I just summarise your proposed changes to ensure I understand them?

You are proposing:
- The SessionId parameter should be a simple string, and not constrained to be a UUID. It is for the API provider to ensure that each SessionId is unique within their own domain and there is no requirement for SessionIds to be unique across all API providers.
- In the 201 Session Created response to a session creation request, and in the 200 Session Info response to a session information request, the "duration" and "startedAt" parameters are removed, but "expiresAt" retained. In other words, the API consumer only gets these responses if the session is already active and "expiresAt" is sometime in the future.

I don't get your point about NotificationData -> QosChanged. A QosChanged schema is defined, but not actually used. Is your proposal to remove this unused schema? I am looking at version 1.2.0 of the API definitions, so it may be that this schema has been used in other versions.

eric-murray:
Thanks @patrice-conil 

So I generally agree with your proposals. But extending notification options beyond the currently specified SESSION_TERMINATED event is a feature extension proposal, and I would suggest to open a new issue to explicitly discuss this.

hdamker:
@patrice-conil is this issue done with #38 and the PR of v0.8.0?

patrice-conil:
@hdamker, yes this problem is solved. I missed the fact that we needed a unique ID across multiple platforms.

PR #67, opened 2022-11-07
Introduce 'qosStatus' and corresponding notification event to fix issue #38

emil-cheung:
A proposal to fix:
https://github.com/camaraproject/QualityOnDemand/issues/38


code/API_definitions/qod-api.yaml
```diff
@@ -329,16 +332,32 @@ components:
...
-        - SESSION_TERMINATED
+      type: object
+      required:
+        - event
+      properties:
+        event:
+          type: string
+          enum:
+            - QOS_STATUS_CHANGED
+        qosStatus:
```
akoshunyadi, line 350: It's little confusing when qosStatus also contains 'reason', a different name like eventInfo would be clearer.

code/API_definitions/qod-api.yaml
```diff
@@ -329,16 +332,32 @@ components:
...
+          type: object
+          properties:
+            status:
+              $ref: "#/components/schemas/QosStatus"
+            reason:
+              type: string
+              enum:
+                - DURATION_EXPIRES
+                - NETWORK_TERMINATES
+                - UNKNOWN
```
akoshunyadi, line 360: I would prefer past tense: DURATION_EXPIRED, NETWORK_TERMINATED

code/API_definitions/qod-api.yaml
```diff
@@ -419,6 +438,17 @@ components:
...
         - severity
         - description
+
+    QosStatus:
+      type: string
+      enum:
+        - REQUESTED
+        - AVAILABLE
+        - UNAVAILABLE
+      description: |
```
akoshunyadi, line 448: I'm not sure that we have the information (event) to set the status to AVAILABLE. Although if so, we could set AVAILABLE immediately on the request and not use REQUESTED.

code/API_definitions/qod-api.yaml
patrice-conil: I also prefer past tense

patrice-conil (dismissed):

code/API_definitions/qod-api.yaml
```diff
@@ -329,16 +332,32 @@ components:
...
-      enum:
-        - SESSION_TERMINATED
+      type: object
+      required:
+        - event
+      properties:
+        event:
+          type: string
+          enum:
+            - QOS_STATUS_CHANGED
```
jlurien, line 349: If `type` is added to support future events, and each type of event may have different properties, this should be designed as a polymorphic object defining `type` as a discriminator, or alternatively define a `oneOf` straucture.

code/API_definitions/qod-api.yaml
```diff
@@ -329,16 +332,32 @@ components:
...
+        qosStatus:
+          type: object
+          properties:
+            status:
+              $ref: "#/components/schemas/QosStatus"
```
jlurien, line 354: We may need to distinguish the values in the response and this asynchronous event. For example, `requested` may make sense for the 20x response, but not for the notification event.

code/API_definitions/qod-api.yaml
jlurien: I also prefer see past participles. But these reason values would be appropriate for a  sort of `session_terminated` event, which I think is necessary explicitly. I don't see them matching the proposed values and definitions for `status`

code/API_definitions/qod-api.yaml
```diff
@@ -329,16 +332,32 @@ components:
...
+      type: object
+      required:
+        - event
+      properties:
+        event:
+          type: string
+          enum:
+            - QOS_STATUS_CHANGED
+        qosStatus:
+          type: object
```
jlurien, line 351: Some property oo the object should be required, otherwise {} would be allowed

code/API_definitions/qod-api.yaml
jlurien: If the objective of this new QosStatus is to allow an inmediate response to `createSession` without knowing if the requested QoS can be granted (allowing an asynchronous provision of the QoS), I see the `requested` in the context of the response, but not for the notification event. In that case, a subsequent event may carry values for `available` or `unavailable`. 
Also, `unavailable` is not appropriate for the `createSession` response, because in that case an error would make more sense instead of returning this value along with a session id. 
I have doubs about the use of `available` in the response. A 201 response should imply that the request is fulfilled. If not, a 202 response would make more sense. Or return a more generic 200 and rely on the status.

jlurien (commented):
I don't fully understand the purpose of this new model after reading issues #33 and #38.  The purpose is to allow an asynchronous model for cases when it is not possible to fulfill the request for `createSession` synchronously?

code/API_definitions/qod-api.yaml
eric-murray: @jlurien
I tend to agree with you, however I think 201 is the correct response whether the QoS profile is granted or only requested because in both cases a session is created which can be later queried, modified (if we add PUT or PATCH) or deleted. And given this, the API caller anyway has to check the qosStatus field to determine if the requested QoS has been granted or is still pending.

So it does not change the API caller's workflow to find out that, in fact, the requested QoS could not be granted and they must either accept that or request a different QoS profile. But it is not clear to me in this case if the session persists (can it still be modified or queried when qosStatus is `UNAVAILABLE`?) or the session is immediately and automatically deleted.

I also can't think of a use case where a notification would have qosStatus as `REQUESTED`.

@emil-cheung
When qosStatus is `UNAVAILABLE`, I think it would be useful to give an indication in the `reason` field as to whether this is only temporary (e.g. due to network congestion) or permanent (e.g. because that QoS profile is just not available where the UE is). So maybe values `CONGESTION` and `NOT AVAILABLE IN THIS LOCATION` could be added as valid reasons why the requested QoS profile is not available. This reason would also need to be included in the 201 response when the qosStatus is `UNAVAILABLE`.

jlurien: @eric-murray
I was initially thinking about a session being created once the request is granted, but that would imply having different ids for requests and sessions. To keep things simpler,  we may consider a session lifecycle starting with "requested", and then moving either to "available/active" or to "unavailable/not_granted".  If a provider is able to resolve the request synchronously it could return the result inmediately.

I see a different status for "terminated", for a session that was "active", and that status may be enriched with reasons like time_expired, abnormal_termination, etc. Status "unavailable/not_granted", for session that was "requested" but never got "active", would have a different set of reasons, like the ones you mention.

emil-cheung:
The deck I presented in the WG meeting associated to this PR:
[https://github.com/camaraproject/QualityOnDemand/blob/dev-v0.9.0/documentation/SupportingDocuments/Presentations/QoS%20Status%20and%20Corresponding%20Notification%20Events.pdf](url)

code/API_definitions/qod-api.yaml
```diff
@@ -325,20 +328,54 @@ components:
...
+            eventInfo:
+              type: string
+              enum:
+                - DURATION_EXPIRED
+                - NETWORK_TERMINATED
+
     SessionEvent:
       type: string
       enum:
-        - SESSION_TERMINATED
```
jlurien, line 341: Is there no case for SESSION_TERMINATED anymore? If we say that the session resource is independent from the qosStatus of the session, if may be more accurate to tell that a session is terminated due to the duration expired, than  that the QoS changed to UNAVAILABLE due to it (as you say below   * `UNAVAILABLE` - The requested QoS cannot be provided by the network due to some reason).
This is subtle anyway, but as said before we should double check that all of this is understood from a developer perspective.

emil-cheung:
To the reviewers who haven't conduct the review, could you please review this PR at your earliest convenience so that we can move forward.

Thanks :)

code/API_definitions/qod-api.yaml
```diff
@@ -325,20 +328,54 @@ components:
...
+      type: object
+      required:
+        - eventType
+      properties:
+        eventType:
+          $ref: "#/components/schemas/SessionEventType"
+      discriminator:
+        propertyName: eventType
+        mapping:
+          QOS_STATUS_CHANGED: QosStatusChangedEventReport
```
SfnUser, line 357: I believe this mapping is incorrect. If this yaml is opened in Swagger, it reports 'QosStatusChangedEventReport' is unused

code/API_definitions/qod-api.yaml
```diff
@@ -174,7 +174,7 @@ paths:
       description: |
         Important: this endpoint is to be implemented by the API consumer.
         The QoD server will call this endpoint whenever any network related event occurs.
-        Currently only SESSION_TERMINATED event is implemented. Any other network events are ignored.
+        Currently only QOS_STATUS_CHANGED event is implemented.
```
eric-murray, line 177: I know I'm being a pedant here, but the OAS file is an API definition, not an implementation. So the description should read:
`Currently only the QOS_STATUS_CHANGED event is defined.`

eric-murray (dismissed):

code/API_definitions/qod-api.yaml
```diff
@@ -325,20 +331,54 @@ components:
...
+      type: object
+      required:
+        - eventType
+      properties:
+        eventType:
+          $ref: "#/components/schemas/SessionEventType"
+      discriminator:
+        propertyName: eventType
+        mapping:
+          QOS_STATUS_CHANGED: "#/components/schemas/QosStatusChangedEventReport"
```
SfnUser, line 360: I believe this is still incorrect. Swagger still reports QosStatusChangedEventReport is unused. Have you tried generating code for this definition? EventReport is empty.

I would simplify this definition with something like below. Please feel free update as appropriate.

```
    EventReport:
      type: object
      required:
        - eventType
      properties:
        eventType:
          $ref: "#/components/schemas/SessionEventType"
        eventValue:
          $ref: "#/components/schemas/QosStatusChangedEventReport"
      
    QosStatusChangedEventReport:
      type: object
      required:
        - qosStatus
      properties:
        qosStatus:
          $ref: "#/components/schemas/QosStatus"
        statusInfo:
          type: string
          enum:
            - DURATION_EXPIRED
            - NETWORK_TERMINATED

    SessionEventType:
      type: string
      enum:
        - QOS_STATUS_CHANGED
```

For this definition, the json schema would be something like below (e.g.):

```
{
  "sessionId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "eventReports": [
    {
      "eventType": "QOS_STATUS_CHANGED",
      "eventValue": {
        "qosStatus": "AVAILABLE",
        "statusInfo": "DURATION_EXPIRED"
      }
    }
  ]
}
```

jpengar:
>To the reviewers who haven't conduct the review, could you please review this PR at your earliest convenience so that we can move forward.
Thanks :)

@emil-cheung I see you have already resolved all comments from last @jlurien review. But actually, some of them were open to discussion. In any case, from TEF side, we are internally reviewing this proposal with Product/Network teams and we will come back with the results of our analysis.

bigludo7 (commented):
Hello
Reviewed with @patrice-conil 

Few comments:

- It will be probably helpful to get the session status in the notification/eventReport to avoid listener to do a GET
- You have introduced use of discriminator for event report (line 357). We get the idea but do you think it's a bit premature as we have only one kind of event?

Thanks
Ludovic/Patrice

emil-cheung:
> 



> Hello Reviewed with @patrice-conil
> 
> Few comments:
> 
> * It will be probably helpful to get the session status in the notification/eventReport to avoid listener to do a GET
> * You have introduced use of discriminator for event report (line 357). We get the idea but do you think it's a bit premature as we have only one kind of event?
> 
> Thanks Ludovic/Patrice

For the point #1: 'qosStatus' is also in 'SessionInfo' datatype, which is also applicable to GET /sessions operation.
For the point #2:  somebody proposed to use discriminator previously. For me, I am open. Let our community to vote for it.

emil-cheung:
> > To the reviewers who haven't conduct the review, could you please review this PR at your earliest convenience so that we can move forward.
> > Thanks :)
> 
> @emil-cheung I see you have already resolved all comments from last @jlurien review. But actually, some of them were open to discussion. In any case, from TEF side, we are internally reviewing this proposal with Product/Network teams and we will come back with the results of our analysis.

Sure. Please free feel to reopen the comments if you still want to further discuss.

code/API_definitions/qod-api.yaml
```diff
@@ -348,31 +348,28 @@ components:
           description: Contains the reported event and applicable information
     
     EventReport:
-      type: object
-      required:
-        - eventType
-      properties:
-        eventType:
-          $ref: "#/components/schemas/SessionEventType"
+      oneOf:
```
patrice-conil, line 351: What's the idea behind usage of discriminator and oneOf with only one ref ?
It might make sense if we extract and common this endpoint for all Camara APIs. Is it the way we follow?

jlurien: I guess this makes sense as long as we foresee new values for eventType for in the future. We can leave this way, prepared for extension, or add the extension syntax when new schemas for other event reports appear

code/API_definitions/qod-api.yaml
jlurien: @SfnUser for me it works with https://editor.swagger.io/#/, which validator are you using? Some of them don't deal well with oneOf semantics.
I think your proposal wouldn't work if we have other event types where  `eventValues` have different schemas. As commented before, we are not still there, so we can decide to simplify it for now and add the extension mechanism when needed.

code/API_definitions/qod-api.yaml
```diff
@@ -325,20 +331,51 @@ components:
...
+      type: object
+      required:
+        - eventType
+        - qosStatus
+      properties:
+        eventType:
+          $ref: "#/components/schemas/SessionEventType"
+        qosStatus:
+          $ref: "#/components/schemas/QosStatus"
+        statusInfo:
```
jlurien, line 368: We should add a description here as well, as we have with QosStatus enum. In this case I would remark that these values only make sense when qosStatus = UNAVAILABLE

code/API_definitions/qod-api.yaml
```diff
@@ -325,20 +331,52 @@ components:
...
+          type: string
+          enum:
+            - DURATION_EXPIRED
+            - NETWORK_TERMINATED
+          description: Currently statusInfo is only applicable when qosStatus is 'UNAVAILABLE'.
```
jlurien, line 373: With this proposal we are assuming that once QoS becomes UNAVAILABLE (due to duration expires or for network reasons), the session is implicitly terminated. I'm wondering if that will be always the case, or there may be situations where QoS is temporarily unavailable or degraded, but session is still active, so QoS could be eventually restored. Is this scenario possible?

code/API_definitions/qod-api.yaml
SfnUser: @jlurien When I made the above comment on Dec 7th, the `oneOf` semantic was not added and hence the error report was valid. However, Emil seem to have fixed that on Dec 8th with this commit https://github.com/camaraproject/QualityOnDemand/pull/67/commits/6d698c02e0050a8205f9474a41350e7599333f52 by adding `oneOf`.

The weird thing is my comment appears on that commit as well (committed on Dec 8th) but my comment is made on Dec 7th. Don't know why.

I agree that with `oneOf` addition, it works OK for me too. Cheers.

hdamker: @SfnUser Thanks for checking again. Github can not automatically evaluating if a comment is addressed by a commit and therefore keeps it. So we can resolve this conversation.

emil-cheung:
@hdamker shall I rebase this PR to the main branch?

hdamker:
> shall I rebase this PR to the main branch?

@emil-cheung sorry for the late answer (thought I had done it already): yes, please rebase the PR to main branch as this is now the branch for the latest changes.

emil-cheung:
> > shall I rebase this PR to the main branch?
> 
> @emil-cheung sorry for the late answer (thought I had done it already): yes, please rebase the PR to main branch as this is now the branch for the latest changes.

@hdamker I have changed the base branch from dev-0.9.0 to main, please check.

hdamker:
> I have changed the base branch from dev-0.9.0 to main, please check.

@emil-cheung Thanks a lot! LGTM, we are ready for a final review by those who have previously approved.

hdamker (approved):

hdamker:
@jlurien @jlurien @eric-murray Could you have a final view?

code/API_definitions/qod-api.yaml
```diff
@@ -352,20 +358,52 @@ components:
...
```
jlurien, line 400:
```suggestion
          description: |
            Reason for the new `qosStatus`. Currently `statusInfo` is only applicable when `osStatus` is 'UNAVAILABLE'.
            * `DURATION_EXPIRED` - Session terminated due to requested duration expired
            * `NETWORK_TERMINATED` - Network terminated the session before the requested duration expired
```

code/API_definitions/qod-api.yaml
```diff
@@ -352,20 +358,52 @@ components:
...
+      type: object
+      required:
+        - eventType
+        - qosStatus
+      properties:
+        eventType:
+          $ref: "#/components/schemas/SessionEventType"
+        qosStatus:
+          $ref: "#/components/schemas/QosStatus"
+        statusInfo:
```
jlurien, line 395:
```suggestion
        statusInfo:
          $ref: "#/components/schemas/StatusInfo"
          
    StatusInfo:
```

As all other properties has its own schema separated

jlurien (commented):
Looks good. A couple of cosmetic suggestions

SfnUser (approved):

code/API_definitions/qod-api.yaml
```diff
@@ -397,7 +397,10 @@ components:
           enum:
             - DURATION_EXPIRED
             - NETWORK_TERMINATED
-          description: Currently statusInfo is only applicable when qosStatus is 'UNAVAILABLE'.
+          description: |
+            Reason for the new `qosStatus`. Currently `statusInfo` is only applicable when `osStatus` is 'UNAVAILABLE'.
```
eric-murray, line 404: typo: \`osStatus\` should be \`qosStatus\`

SfnUser (approved):

eric-murray (approved):

hdamker (approved):

Issue #31, opened 2022-07-29
Ensure secure handling of callback URL


shilpa-padgaonkar:

akoshunyadi:
To solve the issue about the unsecure callback I propose to use oauth2 instead (!) of the current api-key.
It means we would need new parameters (maybe in a substructure): Client-id, Client-secret and Access token URL.
Any comments?

jlurien:
There is an [open issue](https://github.com/camaraproject/WorkingGroups/issues/156) in Commonalities to discuss the Subscription/Notification model across all APIs, and security aspects is also something we should probably agree on.

akoshunyadi:
We could also keep the current api-key and add the additional security. Currently the notification content is really not so important, but I'm thinking of future events also in other APIs with sensible data.

chrishowell:
I think mandating all notification recipients implement OAuth flows is a super bad idea, it's way outside the resources that some developers have available to them.

It would be nice to have centralised management of notification servers though, perhaps each having a different set of access credentials, which could be OAuth, or just Bearer. I think this belongs in Commonalities though, rather than QoD.

hdamker:
@chrishowell yes, discussion is happening in Commonalities and everyone is encouraged to support there with reviews and contributions. We will continue here when the guidelines are updated.

hdamker:
This issue is be done, current v0.10.0-rc is following the design guidelines for notifications as defined by Commonalities v0.2.0. If there are additional security requirements by Commonalities we need to open a new issue.

PR #69, opened 2022-11-10
Add apikey to all endpoints

PhilRW:
We are using the Flask [Connexion](https://connexion.readthedocs.io/en/latest/) framework to develop a demo using this API, but it does not function properly without a security implementation defined for all endpoints.

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -115,6 +115,8 @@ paths:
                 message: "Session could not be created"
         "503":
           $ref: "#/components/responses/Generic503"
+      security:
+        - apiKey: []
   /sessions/{sessionId}:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


hdamker (commented):
@PhilRW: The security schema for all endpoints is already defined in line 14/15:

```
security:
  - oAuth2ClientCredentials: []
```
This should be only `OAuth2ClientCredentials` in accordance with the agreed AuthN and AuthZ concept: https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/CAMARA-AuthN-AuthZ-Concept.md

The `apiKey` security schema is currently in the YAML only for notifications to the callback URL and will be deleted with the PR for issue #31. It should not be used for the API endpoints.

PhilRW:
Okay, thanks. I will update our code to utilize oauth2.

PR #73, opened 2022-12-01
Reduce to one example QoSProfile_Mapping_Table.md

### LINKED ISSUE #72: Proposal to reduce QoSProfile_Mapping_Table.md back to one example
There was a discussion about the changes in [QoSProfile_Mapping_Table.md](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/API_documentation/QoSProfile_Mapping_Table.md) in our last call, which were at that point of time already merged. The discussion was mainly about second example table which maps the same 4 QoS profiles from the current API spec to very different QoS characteristics.

The two main points received also in bilateral discussions were:

- A QoS profile label should map in principle to the same connectivity characteristics across different communication service provider implementations. That does not exclude that service levels, conditions and restrictions will be still subject to an agreement between API invoker and the communication service provider. In this sense the table can only be an example and that’s expressed, I suggest.
- There were several voices who asked not to use the same profile labels for “non-guaranteed” QoS profiles and for “guaranteed” (in the text: “statistically assured”) QoS profiles. 

One proposal was to use different labels for the QoS profiles described in the second table. I suggest that it is too early now to extend already the list of QoS Profile labels (we rather have implementations for the ones defined in v0.8.0).

The proposal would be therefore to leave it for now with the four QoS Profile labels defined in v0.8.0 of the API definition, as it is too early to agree already on additional QoS profile labels. That would mean also to restrict [QoSProfile_Mapping_Table.md](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/API_documentation/QoSProfile_Mapping_Table.md) again to one (the first) example table. Additional QoS Profile labels and it’s exemplary mapping can then be added together with following versions of the API.

Please reply to the discussion on the mailing list, comment here or on the related PR.

hdamker:
Will fix #72 if merged - reduce back to one example in QoS Profile Mapping Table. See the issue for background


jlurien (commented):
I would like to suggest these changes to the remaining table, to be more aligned with short term implementions:

| QOS\_L   | The 5G System throughput is prioritized up to certain higher limit (e.g. 20Mbps), or without an explicit limit. At high load, the throughput may be capped at the limit and can be degraded to lower effective throughput (without a minimum level).  | The service is only provided, when the UE is located within the home network.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
| QOS\_M   | The 5G System throughput is prioritized up to certain medium limit (e.g. 8Mbps). At high load, the throughput may be capped at the limit and can be degraded to lower effective throughput (without a minimum level). | The service is only provided, when the UE is located within the home network.<br>The access to the network service is restricted to a maximal number of simultaneous session. |
| QOS\_S   | The 5G System throughput is prioritized up to certain lower limit (e.g. 4Mbps). At high load, the throughput may be capped at the limit and can be degraded to lower effective throughput (without a minimum level). | The service is only provided, when the UE is located within the home network.<br>The access to the network service is restricted to a maximal number of simultaneous session. |

hdamker:
> I would like to suggest these changes to the remaining table, to be more aligned with short term implementions

@jlurien: proposal is fine for me, have adapted it in https://github.com/camaraproject/QualityOnDemand/pull/73/commits/a3b2a3f977702a8ea0494585c71db7778739d4c7. I also changed QoS_E in similar manner, please have a look.

hdamker:
Will merge based on the agreement within our call on Friday Dec 2nd and the final review by Jose/Jesus.

PR #89, opened 2023-01-16
Update from notificationsUri to notificationsUrl

### LINKED ISSUE #88: Consistent usage of notificationUri vs notificationUrl
# Description of the bug
## notificationsUri vs notificationsUrl
In https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_definitions/qod-api.yaml#L43 a parameter is called `notificationUrl`. In https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_definitions/qod-api.yaml#L268 it is called `notificationUri`. They should use the same name name `notificationUrl`.

## Wrong example of notificationsUrl
The example should also just contain the server reference without the `/notifications`-endpoint (see https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_definitions/qod-api.yaml#L271)

### Why?
The code-generator will generate a class `SessionNotificationsCallbackApi`. 
Inside of this class it is already defined, that the endpoint `/notifications` will be executed.
It will look like: 
`apiClient.invokeAPI("/notifications", HttpMethod.POST, Collections.<String, Object>emptyMap(), queryParams, postBody, headerParams, cookieParams, formParams, localVarAccept, localVarContentType, authNames, returnType);`

The ApiClient will use the 'notificationUrl' from the notificationRequest here: https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_code/core/src/main/java/com/camara/qod/service/QodServiceImpl.java#L265
 If the notificationUrl of the request will now be "https://application-server.com/notifications" then the application will post a notification against `https://application-server.com/notifications/notifications` 

This Issue is an updated version of [Issue 23](https://github.com/camaraproject/QualityOnDemand/issues/23).

maxl2287:
fixes #88

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -265,10 +265,10 @@ components:
           $ref: "#/components/schemas/PortsSpec"
         qos:
           $ref: "#/components/schemas/QosProfile"
-        notificationUri:
+        notificationUrl:
           type: string
           format: uri
-          example: "https://application-server.com/notifications"
+          example: "https://application-server.com"
           description: Allows asynchronous delivery of session related events
         notificationAuthToken:
           type: string
--- a/documentation/API_documentation/QoD_API.md
+++ b/documentation/API_documentation/QoD_API.md
@@ -148,7 +148,7 @@ Following table defines API endpoints of exposed REST based for QoD management o
             <td>Qualifier for the requested latency/throughput profile, e.g. QOS_E</td>
         </tr>
         <tr>
-            <td><b>notificationUri (optional)</b></td>
+            <td><b>notificationUrl (optional)</b></td>
             <td>URI of the callback receiver. Allows asynchronous delivery of session related events, e.g. 'https://application-server.com/notifications'</td>
         </tr>
         <tr>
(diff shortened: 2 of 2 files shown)


hdamker:
LGTM, at least one further review from @eric-murray @jlurien or @patrice-conil would be good.

patrice-conil (dismissed):

eric-murray:
I had a couple of questions / comments on this PR:
- Why are modifications being proposed for the "legacy" implementation within the main QualityOnDemand repository rather than moving the implementation to the new QualityOnDemand_PI1 repository? I had understood that implementations would no longer be stored within the main API repository.
- A change is being proposed to a copy of v0.1.0 stored within the implementation, but no change is being proposed to the version number **of that copy**. This is potentially confusing if such files are viewed out of context. I'd prefer if modifications made for the purposes of an implementation have a suitably extended version number. I'd propose a company specific extension to the base version number (e.g. 0.1.0-dtag.1) when such changes are made. This is the approach that we took within Vodafone when modifying the API specification for our specific implementation (e.g. update of apiRoot and addition of OAuth scope).

jlurien:
I'm OK with renaming parameter to `notificationUrl`,  but I agree with @eric-murray. I think we should fix code/API_definitions/qod-api.yaml and doc here, and move implementations to the new repos.  

We could upgrade yaml version to 0.8,1 and take advantage of the change to incorporate the other typos identified.

maxl2287:
@jlurien and @eric-murray thanks for your feedback.
I will then revert the implementation changes and just change the qod-api.yaml.

hdamker:
@maxl2287 please keep also the change in documentation. And I recommend not to touch the version number in the spec file until we decide to make a new release (like v0.8.1).

jlurien:
> @maxl2287 please keep also the change in documentation. And I recommend not to touch the version number in the spec file until we decide to make a new release (like v0.8.1).

@hdamker Should we resolve the release first and then apply this change towards 0.8.1?

maxl2287:
Branch is now updated without implementation changes

eric-murray:
> @maxl2287 please keep also the change in documentation. And I recommend not to touch the version number in the spec file until we decide to make a new release (like v0.8.1).

@hdamker Will you create the release v0.8.0 first before merging this PR? Although the change is small, our current implementation uses the current v0.8.0 and not this update, and anyone looking for v0.8.0 in the github should find the current version and not subsequent updates that have not incremented the version number.

hdamker:
> Will you create the release v0.8.0 first before merging this PR?
@eric-murray Yes, will do today. I have exactly the same intention to get the v0.8.0 release before any further fix merged.

eric-murray (approved):

hdamker:
@jlurien @patrice-conil can you approve (again)? Then we should be ready here.

shilpa-padgaonkar (approved):

jlurien (approved):
LGTM

PR #94, opened 2023-01-17
Add-501-error-code-[NOT_IMPLEMENTED]-main (#4)

### LINKED ISSUE #81: Introducing error code 501 "Not Implemented"
In current spec the error code 400 ("Bad Request") is used also for cases where the request is correct according to the spec but the implementation is not supporting certain parameter or is expecting other parameters then the API caller has provided.

We propose to introduce error code 501 "Not implemented" e.g. for valid UE identifiers which are not implemented.

ravindrapalaskar17:
Refer to the issue (“Will fix #81”).

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -113,6 +113,15 @@ paths:
               example:
                 code: INTERNAL
                 message: "Session could not be created"
+        "501":
+          description: Not Implemented
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+              example:
+                code: NOT_IMPLEMENTED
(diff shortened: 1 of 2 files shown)


ilya-smirnov-berlin:
Since this feature will be used in 0.9.0 I suggest to set the version to 0.9.0

hdamker:
@ilya-smirnov-berlin thanks for clarification. As the https://github.com/camaraproject/WorkingGroups/pull/123 is currently we will need the PR after our fix release against main. There will be no longer a dev branch, latest code will be in main. Order of PRs will be coordinated in our sub project calls.

For now I will close the PR ... to be reopen after the 0.8.1 release.

jlurien:
I have just submitted a new PR #104 reformatting errors according to the guidelines. We should review and merge the other one first and then align this one to the new format.

Issue #97, opened 2023-01-19
Alignment with API design guidelines


jlurien:
Some weeks ago, a set of [API design guidelines](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md) have been approved in commonalities.

Regarding terminolgy and style:

- Reduce telco-specific terminology in API definitions. 
  - This is treated in issue [#90](https://github.com/camaraproject/QualityOnDemand/issues/90).
- URI paths in lowercase and separated-by-hyphens.
  - Our paths are single word and already in lowercase
- OperationIds are defined in lowerCamelCase.
  - We are already compliant
- Objects are defined in CamelCase inside properties field. This affects declarations within definitions that are $ref in other parts.
  - We are already compliant
- API input and output business data will follow the snake_case notation. 
  - **We are NOT compliant**. This affects property names.
  
Errors must include:

- A field `status`, which can be identified in the response as a standard code from list of Hypertext Transfer Protocol (HTTP) response status codes. 
  - **We have to add it to our model**.
- A unique error `code`, which can be identified and traced for more details. It must be human readable; therefore, it must not be a numeric code. In turn, to achieve a better location of the error, you can reference the value or field that is causing it, and include it in the message.
  - We should review our current codes and probably define new codes for expected logical errors.
- A detailed description of `message`.
  - We are already compliant.

## Proposed actions

- Most of changes apply to current property names, as they must be formattted in snake_case. This can be addressed together with issue [#90](https://github.com/camaraproject/QualityOnDemand/issues/90) in a new PR, when new terms are agreed.

- We have to adapt errors, adding `status`, and at a same time we could enhance current documentation and examples. This can be done in separate PR.

hdamker:
All done.

PR #104, opened 2023-01-25
Align error format with Commonalities guidelines

jlurien:
As discussed in issue [#97](https://github.com/camaraproject/QualityOnDemand/issues/97), we have to align current error model and examples to the guidelines:

- Add new `status` property
- Created new `Generic500` response and $ref all previous instances


code/API_definitions/qod-api.yaml
```diff
@@ -481,14 +469,26 @@ components:
...
+    Generic500:
+      description: Server error
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: "500"
+            code: INTERNAL
+            message: "Session could not be deleted"
```
SfnUser, line 484: This message is not generic to all the procedures.

jlurien: You are right. In the generic model, example should show a more neutral message, I will correct it. Implementations may adapt the text to the scenario, as with other errors.

jlurien, line 484:
```suggestion
            message: "Internal server error: ..."
```

hdamker:
@jlurien Short question: is this PR intended for the minor fix release (which I want to close "now") or already for v0.9? As status is a required field, it might fit better into v0.9.0?

jlurien:
> @jlurien Short question: is this PR intended for the minor fix release (which I want to close "now") or already for v0.9? As status is a required field, it might fit better into v0.9.0?

0.9 definitely

SfnUser (approved):

hdamker:
Added rewievers @eric-murray @patrice-conil as discussed within our call. @SfnUser please renew your review as well. Thanks!

patrice-conil:
Hi @jlurien,
Just a sample to illustrate my review
ErrorInfo definition may look like this:

 ErrorInfo:
      discriminator: type
      type: object
      required:
        - type
        - status
        - code
        - message
      properties:
        type:
          description: The error class name
          type: "string"
        status:
          type: string
          pattern: "^[4-5][0-9]{2}$"
          description: HTTP status code returned along with this error response
        code:
          type: string
          description: Code given to this error
        message:
          type: string
          description: Detailed error description

jlurien:
> Hi @jlurien, Just a sample to illustrate my review ErrorInfo definition may look like this:
> 
> ErrorInfo: discriminator: type type: object required: - type - status - code - message properties: type: description: The error class name type: "string" status: type: string pattern: "^[4-5][0-9]{2}$" description: HTTP status code returned along with this error response code: type: string description: Code given to this error message: type: string description: Detailed error description

Hi  @patrice-conil, have you commented in the file? I guess this is about the problem you mentioned in the meeting with `allOf` for some openapi generator. `ErrorModel` is defined in the commonalities so any change should be aligned there. But in case we decide to define a discriminator to the base class, we could use the existent `status` or `code` for this, along with a `mapping`. Adding a new required property `type` is would repeat the same information.

Another alternative, as we are not reusing that much of the base `ErrorModel`, it is just to add the message property to all GenericXXXResponses and remove the `allOf`. I would not be surprised if some tool has problems as well with OAS3 `mapping` as that is not supported in legacy Swagger 2.

patrice-conil:
> 

Hi @jlurien, yes I commented in the file. and it is indeed linked to the problem that I mentioned last Friday. I agree with you, this mapping should do the trick... but it's quite new and probably not so well supported by generators.
I think simplifying the generation of client side is one way to have a usable API. Perhaps we should spend more time documenting our validation rules than documenting sample responses to describe the behaviour of our proposal.

code/API_definitions/qod-api.yaml
```diff
@@ -452,43 +437,112 @@ components:
...
+                    type: string
+                    enum: ["403"]
+                  code:
+                    type: string
+                    enum: ["FORBIDDEN"]
           example:
+            status: "403"
             code: FORBIDDEN
             message: "Operation not allowed: ..."
     SessionNotFound404:
```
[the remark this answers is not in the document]
jlurien: I could have named it Generic404 as well, but as the only case in this API for 404 was Session Not Found, I chose a more specific name and added an explicit message in the example. Problem with Generic responses is that example has to be generic as well. For automatic API publishing in a web portal it is convenient to embed good examples in the spec, but for client generation purposes it may complicate things. 

I can change the name to `Generic404` if that is more coherent with the rest of responses

jlurien:
> > 
> 
> Hi @jlurien, yes I commented in the file. and it is indeed linked to the problem that I mentioned last Friday. I agree with you, this mapping should do the trick... but it's quite new and probably not so well supported by generators. I think simplifying the generation of client side is one way to have a usable API. Perhaps we should spend more time documenting our validation rules than documenting sample responses to describe the behaviour of our proposal.

I only see the latest comment about Generic404. If mapping is not well supported, I prefer to remove the `allOf` that adding an additional property `@type` in the API spec for all errors to overcome the limitation, as that would add too much redundancy in the response:
```
HTTP 404
{ 
  "@type": "Generic404",
  "status": "404"
  "code": "NOT_FOUND",
  "message": "Session not found"
}
```

code/API_definitions/qod-api.yaml
```diff
@@ -369,9 +349,14 @@ components:
     ErrorInfo:
       type: object
       required:
+        - status
         - code
         - message
       properties:
+        status:
+          type: string
+          pattern: "^[4-5][0-9]{2}$"
```
eric-murray, line 358: I don't see the value of including a pattern in a response object. Are we expecting the client to error check our error responses? Status should really be an integer anyway, but if it is to be a string, the client doesn't need to know any more than that.

code/API_definitions/qod-api.yaml
```diff
@@ -452,43 +437,112 @@ components:
...
+            allOf: 
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties: 
+                  status:
+                    type: string
+                    enum: ["400"]
+                  code:
+                    type: string
+                    enum: ["INVALID_INPUT"]
```
eric-murray, line 449: My understanding is that the point of this `allOf` specialisation is to link the `status` (in this case, `"400"`) with the `code` (`"INVALID_INPUT"`). I don't see the value in this, and would prefer just to give a comprehensive list of examples, differing only by `message` for each of the response schema (`Generic400`, etc.).

If the status and code of a given response received by the client do not align (e.g. `status` is `"404"` and `code` is `"INVALID_INPUT"`), I don't think the client would need this specialisation to detect that the server implementation was broken. So my preference would be simplify life for the client developer reading the API specification and have all error responses being simple `ErrorInfo` objects with no specialisation but a good list of examples they might receive.

SfnUser: Given that this change is triggered from [Api design guidelines](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md) would it be better if we extract these common error messages and add it to commonalities repo - we are moving in that direction looking [here](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/CAMARA_common.json)? We could later reference the same in QoD spec. Of course there will be more open items like versioning spec across repos.

The discussion will be wider and the `common` elements will be binding on all sub-projects to include it in their respective specs. My humble opinion!

eric-murray: Thanks @patrice-conil 

On your specific point:

> If status and code are so related...why do we need both?

There are two reasons:
- HTTP status codes are a standard (primarily IETF RFC 9110) that is well known to developers, whereas CAMARA error codes (`"INVALID_INPUT"`, etc.) are not a standard (albeit that they are fairly intuitive)
- There is not necessarily a one-to-one relationship between `status` and `code`. For example, I think that both `"INVALID_INPUT"` and `"OUT_OF_RANGE"` are `"400"` errors.

So both are useful.

patrice-conil: Thanks @eric-murray,
Good point.
Everyday I learn from my mistakes ... and I learn a lot. :D

code/API_definitions/qod-api.yaml
```diff
@@ -452,43 +437,112 @@ components:
...
+            allOf: 
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties: 
+                  status:
+                    type: string
+                    enum: ["500"]
+                  code:
+                    type: string
+                    enum: ["INTERNAL"]
```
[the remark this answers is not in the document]
hdamker: "Internal Server Error" is also the description for the status code "500" within [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). Therefore +1 .

code/API_definitions/qod-api.yaml
hdamker: This one is done ... will resolve the conversation.

code/API_definitions/qod-api.yaml
jlurien: In the guidelines it appears as string but it could be an integer as well to me. Regarding patterns, we have patterns in other objects returned in responses. To me, it guides the implementation and enforces them to restrict values. 
All of this should be reflected in the guidelines so all APIs in the family design error responses in the same way.

maxl2287: Two comments from my side: 

### 1. String or Integer

 @eric-murray: Based on the API design guideline for ErrorResponses, the status is expected as a String. 
``` json
{
   "status": "400",
   "code": "INVALID_ARGUMENT",
   "message": "A human readable description of what the event represent"
}
```
(see: https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#6-error-responses)

otherwise it would be with integer a example like that:

``` json
{
   "status": 400,
   "code": "INVALID_ARGUMENT",
   "message": "A human readable description of what the event represent"
}
```

### 2. Pattern

@jlurien: I agree that a specific pattern-definition for Http-Status for 4xx or 5xx is not necessarry / I do not see the value, too.

code/API_definitions/qod-api.yaml
jlurien: I agree with @SfnUser that this is something to take to Commonalities, but probably we are leading the path here.

Regarding whether responses must be restrictive or not, it depends on how far we want to go with OAS3 capabilities.

If we just ask error responses just to comply with ErrorModel, a server returning

```
HTTP 400
{
  "status": "HTTP 400",
  "code": "MY_OWN_CODE",
  "message": "My message"
}
```
would strictly be API compliant. We can encourage implementations to follow guidelines and examples without defining specific schemas per response, and it is true that some tools have problems dealing with complex schemas for auto generation of data models, so there are pros and cons. In the end we have to assure that all implementations follow the same model.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
jlurien: I agree on aligning as much as possible to standard HTTP terms.

code/API_definitions/qod-api.yaml
eric-murray: I agree that a common approach amongst all CAMARA APIs is preferable. But let's resolve the specific issues for this API, and then the template established here can be pushed to Commonalities.

On this issue, there is no right or wrong answer. I prefer a "minimalist" API specification that still allows the typical developer to use the API (*"Make everything as simple as possible, but not simpler"*, attributed to Einstein). But others may prefer to know that converting our status string to a short integer is perfectly reasonable, or that they don't have to worry about exceeding `maxsize` when storing the string. It's personal preference.

eric-murray: > Based on the API design guideline for ErrorResponses, the status is expected as a String.

Sure, that is clear, but doesn't change my point that an integer type would be a better fit for what is, after all, known to be a 3 digit integer. And it could be changed, just not by this WG. Interpreting everything as a string, whilst usually successful, also does not usually help much with the next step of processing that data.

hdamker: RFC 9110 says clearly: [The status code of a response is a three-digit integer code](https://www.rfc-editor.org/rfc/rfc9110.html#name-status-codes:~:text=The%20status%20code%20of%20a%20response%20is%20a%20three%2Ddigit%20integer%20code). My proposal to open an issue at Commonalities for that (I can do that) and go in this PR with integer for the status code. In a sense we are "testing" the guidelines here.

maxl2287: @hdamker: I also agree with Integer for this, but if we allign on that then @jlurien has to update all examples where "status" is marked currently as a string-value.

jlurien: Fine with integer, that ends the discussion for pattern as well, but better to move the discussion to commonalities, so other colleagues who don't follow this WG are aware of it. In the end I'll align the PR with all we agree.

hdamker: Opened https://github.com/camaraproject/WorkingGroups/issues/143 on the discussion String vs Integer. 

But: do we use the right field for the numeric HTTP status code? Shouldn't it go into "code" instead of "status"?

[OpenAPI Spec](https://spec.openapis.org/oas/v3.1.0#response-object): 
```
{
  "components": {
    "schemas": {
      "ErrorModel": {
        "type": "object",
        "required": [
          "message",
          "code"
        ],
        "properties": {
          "message": {
            "type": "string"
          },
          "code": {
            "type": "integer",
            "minimum": 100,
            "maximum": 600
          }
        }
      }, ...
```
So "code" is the HTTP status code as Integer. No "status" here ... but that can be an addition.

Shortly skimmed through the Camara sub projects ... no consistent picture.

jlurien: One is really the `status_code` while the other is thought as the `error_code`, the latest intended to enumerate some logical errors of the API, which could help clients to react better to them, as messages are human friendly but not convenient for programming logic.  For example MSISDN_REQUIRED, IP_ADDRESS_REQUIRED, QOS_PROFILE_NOT_SUPPORTED, etc. So far we are not defining a list of error codes, and in the end we only have one generic error code per status code, so is somehow redundant.

jlurien: It seems that we are going with the integer, so I'll update the PR accordingly.

hdamker: @jlurien: The issue in commonalities has been closed and the guideline changed accordingly (PR https://github.com/camaraproject/WorkingGroups/pull/146). So I suppose, we are good to finalize this PR here?

code/API_definitions/qod-api.yaml
patrice-conil: Thanks @eric-murray, 
It's clear now.

jlurien:
Please review the latest version. Some comments:

- `status` is now an integer
- `code` values aligned with Commonalities (INVALID_ARGUMENT, UNAUTHENTICATED, PERMISSION_DENIED...)
- Specific schemas defined for all errors, so `ErrorInfo` schema is not used. I have created a new issue in Commonalities to decide whether to reference specific model per error or the common model: <https://github.com/camaraproject/WorkingGroups/issues/151>. When a common approach is decided we should align this to it.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
jlurien: I have aligned the codes with the ones in Commonalities, and currently they define INTERNAL for 500 and UNAVAILABLE for 503. `code` has not to be necessarily the description of the HTTP code, but in any case, the discussion for this should be moved to Commonalities.

code/API_definitions/qod-api.yaml
jlurien: @hdamker updated, please see my latest comment.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
hdamker: Ok for now ... we just should get a conclusion on that before we close v0.9.0.

code/API_definitions/qod-api.yaml
hdamker: @jlurien Thanks!

code/API_definitions/qod-api.yaml
hdamker: Guess that's done for now

hdamker:
LGTM, just asking @patrice-conil @maxl2287  @eric-murray for a final review (and closing remarks for the open conversations).

hdamker:
@jlurien would you like to rebase your PR to the updated main branch or should I give it a try to resolve the conflicts (and commit them into your PR)?

eric-murray:
Is it intended to approve and merge this PR before Commonalities [Issue #151](https://github.com/camaraproject/WorkingGroups/issues/151) is resolved?

jlurien:
> Is it intended to approve and merge this PR before Commonalities [Issue #151](https://github.com/camaraproject/WorkingGroups/issues/151) is resolved?

We could merge it as it is now and later update the $refs to the right schemas, and remove the unused ones. Indeed I have intentionally left the ErrorInfo schema even if it is currently unused.

eric-murray (approved):

hdamker (approved):

SfnUser (approved):

PR #116, opened 2023-02-20
fix/restore-QosStatus-schema

jlurien:
#/components/schemas/QosStatus was accidentally removed and it is missing


hdamker (approved):
Thanks, seems that it happen in the merge conflict. Will merge it directly.

Issue #34, opened 2022-08-05
Resolving ueAddr ambiguity when it is the UE private address


eric-murray:
The current QoD API definitions identify the UE by its source IP address (ueAddr). If that address is public (i.e. routable) then knowledge of any UE source port (via the uePorts parameter) will allow the UE to be uniquely identified, albeit that this is not trivial. So if the specified UE source IP address is public, uePorts becomes a mandatory additional parameter.

But if the UE source IP address is private (i.e. non-routable), then it may not be possible to uniquely identify the UE. This is because the private address is allocated by the PDN GW to which the UE is connected, and a typical mobile network will have several of these with the specific PDN GW being selected based on network routing policies. PDN GWs will typically use the same private address space (usually the 10.0.0.0/8 range) which means that it is possible that more than one UE is currently allocated a given private IP address.

For the T8 interface, 3GPP resolve this ambiguity by requiring that an ipDomain parameter is also specified. This parameter resolves the ambiguity by identifying which PDN GW (i.e. which "domain") has allocated the private IP address. But I don't think CAMARA can adopt this solution as the API consumer is unlikely to know this information.

If the API consumer is the UE itself, then other approaches can be used to uniquely identify the UE, such as OpenID connect or the public source IP that will be seen by the API gateway even if not explicitly sent as an API parameter. And when the agreed [proposal](https://github.com/camaraproject/WorkingGroups/issues/4) to allow use of GPSI as a UE identifier is implemented, then that can be specified rather than the UE source IP address.

But if the only identifying parameter for the UE is the private source IP address, then either a mechanism to resolve this ambiguity must be introduced by CAMARA, or the documentation should make clear that specifying a private IP for ueAddr is likely to result in an error response (probably 400 BAD REQUEST). Unfortunately I don't have a good proposal as to how this ambiguity can be resolved, so Vodafone's current position is that we will return an error message if the UE source IP specified is private.

hdamker:
Depending on outcome of the discussion/issue the API documentation need to update (see https://github.com/camaraproject/QualityOnDemand/pull/17#discussion_r933064903)

eric-murray:
We can derive the ipDomain from the UE public IP address if known. This is much simpler than trying to derive the UE identity itself from the public IP + port, but requires the API to know both the private and public UE IP to identify the UE. If it is the UE itself calling the API, then the public IP can be determined from the source IP of that request. But otherwise that parameter needs to be passed explicitly along with the private IP.

[...]

PR #123, opened 2023-02-24
Update UeId definition in qod-api.yaml

eric-murray:
- Separate fields introduced for private or public UE IP address in UeId object
- Require port to be additionally specified if UE is identified by its public IP address
- Require public IP address to be additionally specified if UE is identified by its private IP address
- Clean up of example error messages for 400 errors and introduction of additional error examples

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -62,41 +62,53 @@ paths:
                     status: 400
                     code: INVALID_ARGUMENT
                     message: "Schema validation failed at  ..."
-                MsisdnRequired:
-                  summary: msisdn is required as part of ueId
+                InsufficientProperties:
+                  summary: ueId parameter must have at least one property specified
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: ueId.msisdn"
-                IPv4Required:
-                  summary: ipv4addr is required as part of ueId
+                    message: "Insufficient parameters specified: ueId"
+                InconsistentProperties:
+                  summary: ueId parameters must identify the same UE
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: ueId.ipv4addr"
-                IPRequired:
-                  summary: Some IP address is required as part of ueId
+                    message: "Multiple inconsistent parameters specified: ueId"
+                UnidentifiedUE:
+                  summary: Specified ueId parameters do not identify a UE
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: ueId.ipv4addr or ueId.ipv6addr"
-                UePortsRequired:
-                  summary: uePorts is required
+                    message: "UE cannot be identifed from specified parameters: ueId"
+                InvalidFormat:
+                  summary: Parameter is incorrectly formatted
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: uePorts"
-                QoSRequired:
-                  summary: qos is required
+                    message: "Invalid parameter format: ueId.msisdn"
+                MissingParameter:
+                  summary: A specified parameter is missing a required property
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: qos"
+                    message: "Missing parameter: ueId.private_ip4_addr.public_ip_address"
+                ParameterOutOfRange:
+                  summary: Parameter specified is outside of valid range
+                  value:
+                    status: 400
+                    code: OUT_OF_RANGE
+                    message: "Specified parameter out of range: ueId.public_ipv4_addr.port"
+                MissingRequiredParameter:
+                  summary: Parameter is required but not specified
+                  value:
+                    status: 400
+                    code: INVALID_ARGUMENT
+                    message: "Required parameter is missing: qos"
                 UePortsRangesNotAllowed:
                   summary: Ranges at uePorts are not allowed
                   value:
                     status: 400
-                    code: INVALID_ARGUMENT
+                    code: OUT_OF_RANGE
                     message: "Ranges not allowed: uePorts"
         "401":
           $ref: "#/components/responses/Generic401"
@@ -398,8 +410,10 @@ components:
           $ref: "#/components/schemas/ExternalId"
         msisdn:
           $ref: "#/components/schemas/MSISDN"
-        ipv4addr:
-          $ref: "#/components/schemas/Ipv4Addr"
+        public_ipv4_addr:
+          $ref: "#/components/schemas/PublicIpv4Addr"
+        private_ipv4_addr:
+          $ref: "#/components/schemas/PrivateIpv4Addr"
         ipv6addr:
           $ref: "#/components/schemas/Ipv6Addr"
       description: User equipment identifier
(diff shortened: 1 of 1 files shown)


code/API_definitions/qod-api.yaml
```diff
@@ -430,6 +444,44 @@ components:
...
+        - public_address
+        - public_port
+      example:
+        {
+          "public_address": "85.255.235.230",
+          "public_port": 60276
+        }
+    PrivateIpv4Addr:
+      description: |
+        Private (non-routable) source IP address directly allocated to the device, and the public (routable) source IP seen by and identifying the device to the application server. If the device does not have a public IP address, then the public_address value should be specified as the private address as well to indicate this.
```
sfnuser, line 466: If the device does not have a ~~public IP address~~`private IP address`, then...

code/API_definitions/qod-api.yaml
```diff
@@ -430,6 +444,44 @@ components:
...
+          $ref: "#/components/schemas/SingleIpv4Addr"
+        public_port:
+          $ref: "#/components/schemas/Port"
+      required:
+        - public_address
+        - public_port
+      example:
+        {
+          "public_address": "85.255.235.230",
+          "public_port": 60276
```
sfnuser, line 462: In the case when both `public_ipv4_addr` & `private_ipv4_addr` are the same, what should be the value of `public_port`?

hdamker:
@jlurien @patrice-conil @sfnuser Would be great if you can review this PR from your point of view.

sfnuser (approved):

code/API_definitions/qod-api.yaml
jlurien, line 466:
```suggestion
        Private (non-routable) source IP address directly allocated to the device, and the public (routable) source IP seen by and identifying the device to the application server. If the device does have a public IP address, then the public_address value should be specified as the private address as well to indicate this.
```
Is my understanding correct? If the device **has** a public address, then copy the value as a private address as well. With private addresses only it is not always possible to identify the device.  For this scenario the name `private_ipv4_addr` would be misleading. IMO `device_ipv4_addr` or `local_ipv4_addr` would cover better both situations.

jlurien (commented):
I'm confused with the use of private_ipv4_addr in cases where public IPV4 addresses are directly allocated to devices. We need to document well how the 3 IP addresses may be combined.

code/API_definitions/qod-api.yaml
eric-murray: @jlurien

No, that is not how I intended this to work.

It works as follows:

### Identify device by private IP address
**IF** the device has been directly allocated a private IP address
**AND** the API caller wishes to identify the UE by that private address
**THEN** the API caller uses the `private_ipv4_addr` field of `ueId` to pass both private and public (CGNATed) IP addresses

If the private IP address has **NOT** been NATed to a public address (this will not be the normal case), then the private address is copied to the `public_address` field to indicate this. This is the exception documented in the description for `PrivateIpv4Addr`.

So the text is correct

### Identify device by public IP address
**IF** the device has been directly allocated a private IP address which has been CGNATed to a public address **OR** the device has been directly allocated a public IP address (this would be unusual)
**AND** the API caller wishes to identify the device by its public address
**THEN** the API caller uses the `public_ipv4_addr` field of `ueId` to indicate both the `public_address` and `port` allocated to the device (the `port` being required to uniquely identify the device in the case of CGNATing)

jlurien: Thanks Eric for the detailed clarification. This type of explanations would be useful also to be added to the API doc. 

Then, for:

> If the private IP address has NOT been NATed to a public address (this will not be the normal case), then the private address is copied to the public_address field to indicate this. This is the exception documented in the description for PrivateIpv4Addr.

In which scenarios can this happen? Private networks where both device and application server are in the same network? And every device is allocated a distinct private address?

eric-murray: Yes, this scenario would be for private or test networks where the private address allocated to the device is unique (within that network).

So the private address is sufficient to identify the device in this case, but I think making `public_address` required is better because this information will indeed be required for the normal case. Unfortunately, OAS does not support making fields conditionally mandatory - they are either required or optional.

emil-cheung:
When we talk about UE IP address, 'private' or 'public' sometimes are confusing.

First of all, 3GPP NEF AsSessionWithQos API accepts UE IP address assigned by 3GPP Core Network, it may be private or public, depends on NAT is deployed or not.

The Application Server communicates with the UE can observe the UE IP address and port from a IP connection (e.g., a TCP connection). The observed UE IP address shall be public (routable).

When there is not NAT, the assigned IP address and the observed IP address will be the same, and both of them are public.
When there is NAT, the assigned IP address will be private and the observed IP address will be public.

We suggest using terms as below:
- UE-assigned IP address or device-assigned IP address
- UE-observed IP address or device-observed IP address
- UE-observed port or device-observed port

code/API_definitions/qod-api.yaml
sfnuser: If we consider the term routable as ip address that is routable in the internet, then the existing text itself seems to be fine. This means private addr (non-routable) and public addr (routable). With that, my understanding is as below:

- UE behind NAT with a private addr & want to identify using public ip addr -> use `PublicIPv4Addr` object
- UE behind NAT with a private addr & want to identify using private ip addr-> use `PrivateIpv4Addr` object with `public_address` and `private_address` assigned with respective ip addrs.
- UE in a private n/w with a private addr & want to identify using private ip addr itself -> use `PrivateIpv4Addr` object with `public_address` and `private_address` being the same private addr
- UE is assigned a public addr (unusual case) & want to identify using public ip addr ->I thought we still be using the `PrivateIpv4Addr` object with `public_address` and `private_address` being the same public ip addr. This is one anomaly to the 'routable' term and as this is an unusual case, I tend to ignore it. (However, @eric-murray reply above '_Identify device by public IP address_' indicates the use of `PublicIpv4Addr` object for this case -> not sure what should be the value of `public_port` would be in this case? This needs clarification)

From an AppServer implementation perspective, it uses the `PublicIpv4Addr` if populated. If not, it uses the `PrivateIpv4Addr` object to identify the device. Seems OK to me (pending the clarification requested above).

eric-murray: @sfnuser 

> UE is assigned a public addr (unusual case) & want to identify using public ip addr ->I thought we still be using the PrivateIpv4Addr object with public_address and private_address being the same public ip addr. This is one anomaly to the 'routable' term and as this is an unusual case, I tend to ignore it. (However, @eric-murray reply above 'Identify device by public IP address' indicates the use of PublicIpv4Addr object for this case -> not sure what should be the value of public_port would be in this case? This needs clarification)

My thought was that, if a public address is allocated directly to the device, then it will not have a private address at all, and so expecting the API caller to use the `private_ipv4_addr` field to identify the device is somewhat counterintuitive.

On your specific point about also specifying the `public_port`, for this scenario it is indeed not strictly required. Not because the device does not have a public source port, but because it has been allocated ALL public source ports. So any port could be specified. This is different to requiring the `private_address`, as such a device definitely does not have a private address (at least, not one that the mobile network would know about).

Given that `public_port` most definitely is required for the CGNAT case, I decided to make that parameter mandatory. Allowing the API caller not to specify it will cause more problems than it solves. Ultimately, there are three parameters that may be required for IPv4 address - public address, private address and public port. The question is what combination of options do we allow the API caller in specifying these parameters.

My expectation is that allocating public IPv4 addresses directly to the device is an entirely hypothetical scenario. If anyone is aware of commercial networks that still do this, I'd be interested in details of those.

sfnuser: @eric-murray 

> My expectation is that allocating public IPv4 addresses directly to the device is an entirely hypothetical scenario. If anyone is aware of commercial networks that still do this, I'd be interested in details of those.

I agree. We have captured the use case and thought out options here anyways. If this do not suffice, a new issue shall be created at that point and discussed. We do not want to overcomplicate now.

hdamker:
There is a discussion paper from @emil-cheung for the topic, currently in #127. Let's have the discussion tomorrow in our regular call.

eric-murray:
I think the discussion here is getting too complicated for a PR, so I'm closing this one and moving the discussion back to Issue #34. I have made an alternative proposal there. Once we have **agreement** in the issue, I will open a new PR - hopefully for approval and not as an invitation to re-open the discussion.

PR #126, opened 2023-03-03
chore: refer-to-common-ErrorInfo-schema

jlurien:
As agreed in the discussion in Commonalities (https://github.com/camaraproject/WorkingGroups/issues/151), we should adopt the new agreement to refer to the common ErrorInfo schema for every error.

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -54,7 +54,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Error400"
+                $ref: "#/components/schemas/ErrorInfo"
               examples:
                 Generic400:
                   summary: Schema validation failed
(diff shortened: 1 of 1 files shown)


hdamker (changes requested):
There are reference to previous error schemas in two lines: 

- 57: `$ref: "#/components/schemas/Error400" `
- 110: `$ref: "#/components/schemas/Error409"`

which need to be replaced.

Rest looks good to me.

jlurien:
> There are references left to previous error schemas in two lines:
> 
> * 57: `$ref: "#/components/schemas/Error400" `
> * 110: `$ref: "#/components/schemas/Error409"`
> 
> which need to be replaced.
> 
> Rest looks good to me.

Indeed. I missed the inline ones. Corrected. Thanks

hdamker (approved):
LGTM now

shilpa-padgaonkar (approved):

Issue #125, opened 2023-03-01
Adds service provider defined QoS profiles


RandyLevensalor:
The current enum of 4 predefined QoS profiles will be replaced with service provider-defined profiles. 3GPP TS 23.203 Rel-15 supports 21 different QoS Class Identifiers (QCIs), and each QCI can be mapped to multiple QoS profiles. Wireline broadband operators typically deploy 50 or more QoS profiles to a single distribution network to address a combination of enterprise and residential needs.

These profiles will be searchable based on profile and end device identifier attributes. Each QoSProfile will contain attributes that describe the profile along with a required name and unique ID. The only change to the Qod session is replacing QoS with a QosId.

Here is an example QoSProfile payload for a QoS profile based on the 3GPP QCI_1 profile for voice:

```json
{
  id:"123e4567-e89b-12d3-a456-426614174000",
  name:"QCI_1_voice",
  status:"Active",
  minBandwidth:{ value:1 unit:"Mbps"},
  maxBandwidth: { value:1 unit:"Mbps"},
  minDuration: { value:1 unit:"Seconds"},
  maxDuration: { value:24 unit:"Hours"},
  priority: 2.0,
  DelayBudget: { value:100 unit:"ms"},
  packetErrorLossRate: 2 
  direction: "Both"
}
```

The diagram below shows the new endpoints, components, and schemas for the proposed change.

![QoS Profile Design - get-only](https://user-images.githubusercontent.com/1365512/222166989-a7125c61-15d8-453e-a7ae-c2a99aaf5957.svg)

A partial implementation of the above change can be viewed in [this fork](https://github.com/cablelabs/QualityOnDemand/pull/2)

eric-murray:
Thanks @RandyLevensalor 

Can you clarify the scope of the `id` parameter? I assume this is unrelated to the "session id" (currently also called `id`) and identifies the QoS profile only (independent of any device that might be using it, so that all devices requesting "QCI_1_voice" get the same `id`)?

What does a `status` of `Active` mean? That the profile can be requested for a device?

I assume `maxDuration` and `minDuration` are the (statically defined) minimum and maximum durations that can be requested for that profile, and not related to the duration that is requested by the client (other than that a request will be rejected if the requested duration falls outside these bounds).

Is `name` intended to be unique? Would it not be "friendlier" to allow the client to request a QoS profile by `name` rather than `id`? Are both `id` and `name` required?

[...]

PR #138, opened 2023-04-19
Create service provider defined QoS Profile

RandyLevensalor:
Implememnted issue #125


RandyLevensalor:
@eric-murray I think that I've incorporated your feedback into the PR.  But there's a decent chance that I missed something.  Thanks!

jlurien (commented):
Thanks for the PR. Some initial comments from a first review

code/API_definitions/qod-api.yaml
```diff
@@ -204,6 +204,153 @@ paths:
...
+          in: query
+          schema:
+            type: string
+          required: false
+        - name: status
+          in: query
+          schema:
+            $ref: '#/components/schemas/QosProfileStatusEnum'
+          required: false
+        - name: targetMinimumUpstreamRate
```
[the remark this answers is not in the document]
eric-murray: The intended serialisation for these complex objects is not clear to me. For example, if I am searching for a profile that supports a minimum upstream data rate of exactly 10 Mb/s, would the serialisation be:

`GET /qos-profiles?targetMinimumUpstreamRate={"value": "10", "unit": "Mbps"}`

I don't believe OAS defined serialisation in that way, so this should be documented.

Also, syntax for filtering is defined in the [CAMARA design guidelines](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#83-filtering), so we should try to follow those. So I should be able to search for all profiles that have a parameter higher or lower than a specified value, which is when filtering becomes much more useful. e.g.

`GET /qos-profiles?priority.gte=5`

which would mean "return all profiles with a priority greater or equal to 5". But how to do this for complex objects?

Maybe: `GET /qos-profiles?targetMinimumUpstreamRate.gte={"value": "10", "unit": "Mbps"}`

I agree with @jlurien that this could all get complex to implement for complex objects, but first lets be clear about how filtering is meant to work and, if different from the mechanism documented in the design guidelines, why is a different approach being proposed and do we need to update the guidelines?

Personally, I think filtering would be useful here, but it is complicated by the use of complex objects in the API definition.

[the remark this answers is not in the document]
jlurien: Unless we have clear requirements for the convenience or necessity of filters, I would go for a simple GET without filters for the first iteration, so we can focus here on how to define a profile.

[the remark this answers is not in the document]
eric-murray: @RandyLevensalor
As mentioned above, the CAMARA API Design Guidelines have a well-defined syntax for [filtering](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#83-filtering).

So:
`GET /qos-profiles?priority=5`
means return all profiles where `priority` is exactly equal to 5, whereas:
`GET /qos-profiles?priority.gte=5`
means return all profiles where `priority` is greater than or equal to 5

If you do not intend this behaviour, you really need to justify why, for otherwise API implementors will have to implement different filtering implementations for different APIs, which they really don't want to do. That's why we have guidelines.

And if you do intend this behaviour, you need to decide how you want it to work for complex objects according to the [OAS parameter serialisation](https://swagger.io/docs/specification/serialization/) rules. I think you may need to add `style: deepObject` if you want code generators to support this, but even then that only works for equalities. As the design guidelines don't consider filtering on complex object values, these would also require an update.

@jlurien 
I think filtering on `name` (the "contains" variant) and `status` is required (why would I want to see `Inactive` or `Deprecated` profiles?). But these are simple strings and the filtering syntax is clear. The value of filtering on other parameters is less clear and, whilst the filtering syntax for these complex objects is also unclear, I'd agree to remove filtering for those parameters for now.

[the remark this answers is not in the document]
RandyLevensalor: @eric-murray

The example you provided around highlights why I think applying numeric operators to many of the qos profile fields is confusing and ambiguous.

If we were to interpert this as a puerly numeric operator, then `GET /qos-profiles?priority.gte=5` would return QoS Profiles with a priority no higher than 5.  For instance a priority of 6 would be returned and that would have a lower priority than 5.

When I read:

> return all profiles where priority is greater than or equal to 5

I would expect that to return higher priority profiles, but a lower numeric priority value.  So a priority 2.5 would be a higher priority than 5.

What about updating the description of the query to say something like
"The get query will return all profiles to can meet the stated filter criteria."

Then for each attribute, we can provide example of what types of results would be expected.  For instance, priority would be the stated priority or a higher priority.  So it would actually be equivalent to the lte numeric operator.

This doesn't address the issues with complex data types.  I'd like to get to consensus on this first.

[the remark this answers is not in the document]
eric-murray: @RandyLevensalor 
But that is how developers will expect filtering to work. If they ask for all profiles with a priority value of 5 or higher, they will expect to see QoS profiles with priority = 6, but not with priority = 4.

Part of the problem is that you have defined `priority` as `The priority of this QoS Profile`, with no indication as to whether a higher number represents a higher or lower scheduling priority. 3GPP's use of lower numbers for higher scheduling priority is just a convention (albeit a common one) and one with which the API consumer may not be familiar. If it is explained that lower priority numbers mean a higher scheduling priority, then it will be obvious that filtering for priority values less than or equal to the lowest scheduling priority acceptable will return a list suitable QoS profiles.

But expecting `GET /qos-profiles?priority.gte=5` to return QoS profiles with a priority value less than or equal to 5 is so counter-intuitive that it should not be implemented.

Anyway, my proposal to @jlurien was only to filter on the string fields `name` and `status`. That remains my proposal for now.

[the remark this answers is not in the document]
jlurien: @eric-murray  I'm fine with limiting filters. `status` makes sense to discard profiles others than active. About `name`, if that is the friendly unique identifier of the profile, probably makes sense to use it as path parameter instead of the `id`,  as `GET /qos-profiles/{name}`. We may not need both `id` and `name` as identifiers.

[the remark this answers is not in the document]
eric-murray: @jlurien My understanding was that `name` does not need to be unique, but `id` does. So `name` is a (fairly short) free text field that the API provider can use to indicate to the API caller for which use cases a specific profile might be useful. But there could be several similar profiles that are given the same name but differ in only minor ways.

For example, an API provider might include the term `conversational_video` in multiple profile names, and the API caller can find them using:
`GET /qos-profiles/?name~=conversational_video`

An API provider who thinks it important that the API caller know what QCI they are getting for their money can include that in the `name`:
`GET /qos-profiles/?name~=qci4`

So searching on `name` makes sense, but not requiring that it be unique also makes sense.

code/API_definitions/qod-api.yaml
```diff
@@ -279,7 +426,9 @@ components:
         applicationServerPorts:
           $ref: "#/components/schemas/PortsSpec"
         qosProfile:
-          $ref: "#/components/schemas/QosProfile"
+          oneOf:
```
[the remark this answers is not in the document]
eric-murray: @jlurien
I assume the intention is to drop this support for the "legacy" QoS profiles at some stage, so `QosProfile` would be deleted and `qosProfile` would then be of type `QoSProfileId` only. If that is the intention, then we should keep the restriction that `QoSProfileId` be a uuid (I agree with your comment below) and live with this workaround until it is agreed that the current QoS profiles are no longer useful.

If that is not the intention, and the legacy profiles must always be supported, then simply redefining `QoSProfileId` to have `minLength: 5` would achieve that, albeit at the cost of more complex validation.

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
       description: |
         Ports may be specified as a list of ranges or single ports.
+    QoSProfileId:
+      type: string
+      example: 123e4567-e89b-12d3-a456-426614174000
+      description: |
+        Unique ID of the QoS Profile to be used for this session.
+      minLength: 36
+      maxLength: 36
+      format: string
```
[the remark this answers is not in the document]
eric-murray: Agree, should be `format: uuid`.

code/API_definitions/qod-api.yaml
```diff
@@ -204,6 +204,153 @@ paths:
...
+        - name: qos-profile-id
+          in: path
+          description: QoS Profile ID that was obtained from the createProfile operation
+          required: true
+          schema:
+            type: string
+            example: foo-bar
+            description: Unique generated ID for a QoS Profile
+            minLength: 3
+            maxLength: 128
```
eric-murray, line 328: Why `minLength: 3` and `maxLength: 128`?

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
+          description: The maximum length of time that this profile can be deployed.
+        priority:
+          type: number
+          example: 5.0
+          description: |
+            Priority for this session
+          format: float
+          minimum: 0
+          maximum: 10
+          default: 5
```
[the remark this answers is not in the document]
eric-murray: A default value does not make sense here. This should be removed.

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
+          $ref: "#/components/schemas/Duration"
+          description: The shortest time period that is supported this profile can be deployed.
+        maxDuration:
+          $ref: "#/components/schemas/Duration"
+          description: The maximum length of time that this profile can be deployed.
+        priority:
+          type: number
+          example: 5.0
+          description: |
+            Priority for this session
```
eric-murray, line 539: `description: The priority that will be given to the session, where a lower value indicates a higher priority`

documentation/API_documentation/QoD_API.md
```diff
@@ -308,6 +307,177 @@ Following table defines API endpoints of exposed REST based for QoD management o
...
+                If this rate is lower than the current provisioned maximum rate, then the current maximum burst rate is used.<br>
+                <b>targetMinimumDownstreamRate</b>: This is the targeted minimum Downstream rate for this profile.<br>
+                <b>maxDownstreamRate</b>: This is the maximum sustained Downstream rate for this profile. If this is undefined, then the maxDownstream rate will not change.
+                If this rate is lower than the current provisioned maximum sustained rate, then the current rate is used.<br>
+                <b>maxDownstreamBurstRate</b>: This is the maximum burst Downstream rate for this profile. If this is undefined, then the maxDownstream rate will not change.
+                If this rate is lower than the current provisioned maximum rate, then the current maximum burst rate is used.<br>
+                <b>minDuration (optional)</b>: Minimum duration for a session with this QoS Profile<br>
+                <b>maxDuration (optional)</b>: Maximum duration for a session with this QoS Profile<br>
+                <b>priority (optional)</b>: The priority of this QoS Profile<br>
```
eric-murray, line 377: <b>priority (optional)</b>: The priority that will be given to sessions using this QoS profile, where a lower value indicates a higher priority

documentation/API_documentation/QoD_API.md
```diff
@@ -308,6 +307,177 @@ Following table defines API endpoints of exposed REST based for QoD management o
...
```
eric-murray, line 452: <b>priority (optional)</b>:  The priority that will be given to sessions using this QoS profile, where a lower value indicates a higher priority

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
jlurien: To me the [proposal](https://github.com/camaraproject/QualityOnDemand/issues/125#issuecomment-1520183394) by @sfnuser makes sense. I don't see why the current predefined qosProfiles should be treated differently once we have the new discovery functionality in place.

We should also decide if API spec enforce some predefined QoS Profile to be offered by all implementations, and what is the best unique identifier for a profile, a UUID or some more friendly label such as current QOS_X. Problem with UUID is that they are not friendly for predefined profiles.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
jlurien: As commented above, I have doubts about the convenience of having UUID in all cases. If this is to be passed when creating a request, it would be always necessary to discover profiles, to know their UUIDs. It is very related with the other topic about the coexistence of legacy and new profiles.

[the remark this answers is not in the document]
eric-murray: If SCEF / NEF is the network element implementing the QoS profile, then my expectation is that the `id` parameter would be the one passed as `qosReference` to the SCEF / NEF. From that point of view, constraining `id` / `qosReference` to be a uuid would greatly simplify data validation, and I would prefer to avoid having to do any mapping between `id` and some other set of `qosReference` values.

However, I take the point that this is linked to the discussion about the use of `name`, and whether that should be unique, and hence the QoS profile key. `name` should not be constrained to be a uuid. So I'd say if `name` is made the key, but we keep `id` as an optional QoS profile parameter, then we just need to say `id` is a string, and it can be set to be a uuid for each QoS profile if that is convenient.

But I would prefer to avoid passing `name` to the SCEF / NEF as `qosReference`, as `name` should mean something to the API consumer that will not necessarily coincide with the internal naming conventions for `qosReference`. I would expect QoS profiles to have an associated `id`, whether that is formally part of the definition, or only an internal parameter.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
RandyLevensalor: > To me the [proposal](https://github.com/camaraproject/QualityOnDemand/issues/125#issuecomment-1520183394) by @sfnuser makes sense. I don't see why the current predefined qosProfiles should be treated differently once we have the new discovery functionality in place.

I can see going down this path of not using a UUID for the QoS profiles and instead use a name as the id.  We can define this name to be a unique string that matches this regex `^[a-zA-Z0-9_]+$`

The length of the name can be from 1 to 128 characters?

With this, the name would replace the ID as the unique identifier.

I'd also like to add an optional `displayName` attribute as a string that could be easier to read, and can contain spaces and other special characters. This would replace the current name attribute and would not be unique.

Since the QOS_S etc. can be a part of this, I think that all of the other attributes would become optional without default values.  This is due to the fact that QOS_S etc. can be defined by only using the name field.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
RandyLevensalor: @eric-murray @sfnuser @jlurien 

What if we change the Duration to be an int64 that is in microseconds.  This can encompass several years and should be small enough to handle future requirements around jitter.  Conversions would need to be done to make these human readable, but it would allow for a consistent and searchable value across all time fields.

For Rate, that can be simplified to bits per second (bps) as a int64.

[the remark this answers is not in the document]
RandyLevensalor: One more note.  Using a name with some constraints is consistent with the way amazon implemented their EC2 instance types.  EC2 instance sizing is a rough analogy for the QoSProfiles.  https://aws.amazon.com/ec2/instance-types.  So this pattern should be familiar to many application developers.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
RandyLevensalor: One more note. Using a name with some constraints is consistent with the way amazon implemented their EC2 instance types. EC2 instance sizing is a rough analogy for the QoSProfiles. https://aws.amazon.com/ec2/instance-types. So this pattern should be familiar to many application developers.

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
hdamker: > Anyway, my proposal to @jlurien was only to filter on the string fields `name` and `status`. That remains my proposal for now.

I second this proposal. The content of this PR should be to get the QoS Profile structure right. 
Instead of complex searches I would envision a new method which allows a developer to define his connectivity needs and they will get a list of matching profiles (or even only the one which is above and nearest to their needs).

hdamker:
@RandyLevensalor Could you make a list of the topics we need to discuss tomorrow in our call - best with the options we have? (and try to resolve the others before the call, just if possible)

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
+          format: float
+          minimum: 0
+          maximum: 10
+          default: 5
+        packetDelayBudget:
+          $ref: "#/components/schemas/Duration"
+          description: The maximum latency for a packet.
```
eric-murray, line 546: description: The Packet Delay Budget is the maximum allowed one-way latency between the device and the internet gateway of the carrier network. It does not include any delay between the carrier network gateway and the application server. If a packet is delayed by more than the packet delay budget, it is considered to have been dropped, and will count towards the Packet Error Loss Rate. See, for example, 3GPP TS 23.203.

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
+          
+            To be more specific, if we consider the round trip latency values of the 99th
+            percentile of traffic, jitter is the measure of how much these values deviate
+            from the average round trip latency for this percentile. A low jitter value
+            indicates that the latency values are relatively consistent, whereas a high
+            jitter value implies that there is a significant variation in latency times,
+            which can negatively impact the performance of real-time applications such as
+            VoIP, video conferencing, and online gaming.
+        packetErrorLossRate:
+          type: integer
```
eric-murray, line 564: `description: The Packet Error Loss Rate is the fraction of total packets sent that are either not delivered to their destination at all, or experience a delay greater than the allowed Packet Delay Budget`

Not sure why PELR should be expressed as 2^(-N) rather than 10^(-N). 3GPP specified PELRs are all 10^(-N), so having to convert these to powers of 2 is inconvenient. But maybe other standards use powers of 2.

code/API_definitions/qod-api.yaml
```diff
@@ -328,6 +477,166 @@ components:
...
+          format: float
+          minimum: 0
+          maximum: 10
+          default: 5
+        packetDelayBudget:
+          $ref: "#/components/schemas/Duration"
+          description: The maximum latency for a packet.
+        jitter:
+          $ref: "#/components/schemas/Duration"
+          description: |
```
eric-murray, line 549: `description: The jitter requirement is the target maximum 99%ile round-trip packet delay variation, as defined by ITU Y.1540, measured over all acknowledged packets in the session. An acknowledged packet is one for which an acknowledgement of receipt is sent back to the sender (e.g. using the TCP transport protocol).`

RandyLevensalor:
@hdamker I'll get the updated PR posted this evening.

@eric-murray thanks for all of your comments and feedback.  I'm working on incorporating it.

RandyLevensalor:
@hdamker 

Open items to discuss:

1. Behaviour and parameters of querying QoS Profiles.  The latest version is very limited and can be updated in a subsequent release.
2. Is mimumumTargetRate acceptable or would guaranteed be more better.
3. The latest draft eliminates the unique ID in favor of a unique name with minimal constraints.  Is this the approach that we want to take?
4. Descriptions in the OpenAPI spec have been revised.  Which ones need additional clarifications? Do these need to be duplicated in the documents as well?

hdamker:
@RandyLevensalor thanks for the summary of open issues

> 4\. Do these need to be duplicated in the documents as well?
No, see #144

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +414,213 @@ components:
...
+            preferential treatment.
+            The lower value the higher priority.
+            Since not all access networks use the same priority range, so this priority will be
+            scaled to the access network's priority range.
+          format: float
+          minimum: 1
+          maximum: 10
+        packetDelayBudget:
+          $ref: "#/components/schemas/Duration"
+          description: |
```
eric-murray, line 501: This description is too 3GPP centric. And it also doesn't match the 3GPP definition, which is the maximum one-way latency between the device and the core network, so excludes the path outside of the carrier's network. So it is not a commitment to maintain the end-to-end latency below this value.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +414,213 @@ components:
...
+            and it ensures a consistent level of performance for specific services within the network.
+        maxUpstreamRate:
+          $ref: "#/components/schemas/Rate"
+          description: |
+            The maximum best effort data
+            Equal to targetMinimumUpstreamRate when max is the as the target. 
+        maxUpstreamBurstRate:
+          $ref: "#/components/schemas/Rate"
+          description: |
+            The burst rate for best effort data
```
eric-murray, line 454: `description: The maximum amount of data that can be sent during any interval equal to the Packet Delay Budget, above which the latency for the packets in that burst may exceed the specified Packet Delay Budget`

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +414,213 @@ components:
...
+            DOCSIS 3.1 - MAC and Upper Layer Protocols Interface Specification" 
+            and it ensures a consistent level of performance for specific services within the network.
+        maxDownstreamRate:
+          $ref: "#/components/schemas/Rate"
+          description: |
+            The maximum best effort rate
+            Equal to targetMinimumUpstreamRate when max is the as the target. 
+        maxDownstreamBurstRate:
+          $ref: "#/components/schemas/Rate"
+          description: |
```
eric-murray, line 477: `description: The maximum amount of data that can be sent during any interval equal to the Packet Delay Budget, above which the latency for the packets in that burst may exceed the specified Packet Delay Budget`

RandyLevensalor:
@eric-murray @hdamker @jlurien @ToshiWakayama-KDDI 

I've created a new issue to begin the discussion around extending the qos-profile queries.  Please chime in there with any thoughts.  #147

hdamker:
Discussion results from call at 2023-05-05

> 1. Behaviour and parameters of querying QoS Profiles.  The latest version is very limited and can be updated in a subsequent release.

Only simple queries within this PR (and v0.9.0), for further enhancements @RandyLevensalor has already opened #147 

> 2. Is mimumumTargetRate acceptable or would guaranteed be more better.

We keep the term minimumTargetRate. Within mobile networks a value >0 can be interpreted as guaranteed bitrate (but still the technical implementation could be use other mechanisms as well).

> 3. The latest draft eliminates the unique ID in favor of a unique name with minimal constraints.  Is this the approach that we want to take?

We taking the apporach id = name. It would be good to have naming conventions to distinct profiles aligned across operators from operator specific ones going forward.

> 4. Descriptions in the OpenAPI spec have been revised.  Which ones need additional clarifications? Do these need to be duplicated in the documents as well?

No need to duplicate, see #144

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +414,221 @@ components:
...
+        name:
+          $ref: "#/components/schemas/QosProfileName"
+          description: |
+            The name of the QoS profile.  This name is used to identify the profile when creating a session.
+            The name must be unique.
+        status:
+          $ref: "#/components/schemas/QosProfileStatusEnum"
+          description: |
+            The status of the QoS profile.  The status is set to ACTIVE when the profile is created.
+            The status is set to INACTIVE when the profile is deleted.
```
sfnuser, line 430: Are we allowing deletion of a profile dynamically?
As `status` in not a required field in `QosProfile`, should we consider profiles with this field not populated by the provider as implied `active`? I wonder if it is a good idea to make `status` as a required field.

This would help developer GET a profile, check its `active` `status`  before using it rather than `active or empty` status. Thoughts?

sfnuser (commented):
Thanks @RandyLevensalor for the updates. I have completed my review of the spec.

hdamker:
@RandyLevensalor would it be possible to resolve the remaining comments before Friday?

code/API_definitions/qod-api.yaml
RandyLevensalor: Good point.  Let's make this required.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +414,221 @@ components:
       description: |
         Ports may be specified as a list of ranges or single ports.
     QosProfile:
+      description: |
+        Data type with attributes of a QosProfile
+      type: object
+      properties:
+        name:
+          $ref: "#/components/schemas/QosProfileName"
```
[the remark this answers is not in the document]
RandyLevensalor: Agreed

code/API_definitions/qod-api.yaml
RandyLevensalor: I'll remove the end-to-end reference.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +400,224 @@ components:
...
+            The lower value the higher priority.
+            Since not all access networks use the same priority range, so this priority will be
+            scaled to the access network's priority range.
+          format: float
+          minimum: 1
+          maximum: 10
+        packetDelayBudget:
+          $ref: "#/components/schemas/Duration"
+          description: |
+            The packet delay budget is the maximum allowable delay between the device or gateway
```
eric-murray, line 494: Should be `maximum allowable one-way delay`.

documentation/API_documentation/QoSProfile_Mapping_Table.md
```diff
@@ -14,3 +15,115 @@ It is generally assumed that there exists an agreement between the API Invoker a
 
 **Note:**
 The QOS profile labels in the table can be used within Camara for validating the QoD APIs.
+
+## ExtendedQosProfile
```
eric-murray, line 19: Definitions below don't match the current definitions in the API OAS definition.

documentation/API_documentation/QoD_API.md
```diff
@@ -308,6 +307,177 @@ Following table defines API endpoints of exposed REST based for QoD management o
...
+        <tr>
+            <td><b>Request Body Parameters</b></td>
+            <td colspan=2><b>No Request Body Parameters defined</b></td>
+        </tr>
+        <tr>
+            <td rowspan=7><b>Response</b></td>
+            <td><b>201: QoS Profile created</b></td>
+            <td>
+                Response body:<br>
+                <b>id</b>: The identifier of the QoS Profile<br>
```
eric-murray, line 437: Definitions below don't match profile definitions in API OAS specification

RandyLevensalor: > Definitions below don't match profile definitions in API OAS specification

@eric-murray I thought that we were going to get rid of the redundant docs before the release, so I didn't need to update these.  Did I not follow that discussion? It is early for me and I'm not at 100% during our calls.

eric-murray: You can remove these additional documents from the PR if you don't want them to be part of the PR. But I don't see the point in updating them with information that does not match the API definition.

jlurien:
Some additional comments:

* The relationship between the 3 upstream/downstream attributes is difficult to understand. `targetMin*` may reflect the tentative bottom limit (avoiding calling it guaranteed). but what is the difference between `max*` and `maxBurst*` from a developer perspective.
* Some reference to latency may be more dev friendly than "packetDelayBudget" but I guess that this term is more accurate
* I miss an optional property, such as `description` to allow some dev friendly qualitative explanation, e.g,. "Low jitter, low packet losses, low latency.", "Real Time gaming" ... 
* Should we differentiate time units per property? Reusing Duration allows other huge units for jitter or delay, and also tiny units for session duration. Not sure if we should enforce some restrictions

documentation/API_documentation/QoD_API.md
jlurien: I would focus on reviewing the descriptions within the spec, as we are deprecating this other doc, but agree that in that case it is better to keep this file as it was so it doesn't show up in the PR

code/API_definitions/qod-api.yaml
RandyLevensalor: @jlurien I went with @sfnuser suggestion.

> This would help developer GET a profile, check its active status before using it rather than active or empty status. Thoughts?

The point to simplify code to allow for always checking the value.  Though good coding standards would always check for a value is not null before comparing it, even if it is required.

What about setting the default to be ACTIVE for this status and having it as a required field?

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,236 @@ components:
...
+            Priority levels allow efficient resource allocation and ensure optimal performance
+            for various services in each technology, with the highest priority traffic receiving
+            preferential treatment.
+            The lower value the higher priority.
+            Since not all access networks use the same priority range, so this priority will be
+            scaled to the access network's priority range.
+          format: float
+          minimum: 1
+          maximum: 10
+        packetDelayBudget:
```
RandyLevensalor, line 500: @eric-murray If we are going to constrain this to one-way, does that mean that we need to have an upStream and a downStream one?

Could we revert this to being a round-trip delay for the profile and then it's up to the operator to map this two up and down delays that fit within this budget?

@jlurien I'm happy to rename this to latency. 

Does anyone else have a comment on this attribute?

eric-murray: Packet delay budget can be assumed the same for both uplink and downlink.

If you want to specify a round-trip delay, then you would need to rename, as otherwise some will assume that the 3GPP definition of packet delay budget applies without reading the definition.

Also, for reasons discussed in the last meeting, this could not be the end-to-end round-trip latency, as likely the server will be located outside the CSP's network, and the CSP cannot assure QoS over an external network. Also, the concept of round-trip latency to a point in the middle of the flow (the CSP Internet Gateway, for example) might be somewhat counter-intuitive, as no packets will be round-tripping from the IGW.

My preference is to keep packet delay budget as one-way latency from device to CSP IGW, and that number can be doubled if someone wants to compute the contribution of the CSP network to overall end-to-end latency. Keeping the name `packetDelayBudget` instead of `latency` will avoid developers thinking this is the end-to-end round-trip latency because they didn't read the documentation.

RandyLevensalor: US latency is usually longer than DS latency, so I don't think assuming that they are the same really works.

If we are set against defining round trip latency on the access network I'd prefer to split this to be `usPacketDelayBudget` and `dsPacketDelayBudget`.

eric-murray: The *latencies* may be different, but the *budget* can be the same. This is the 3GPP definition:

> The Packet Delay Budget (PDB) defines an upper bound for the time that a packet may be delayed between the UE and the PCEF. For a certain QCI the value of the PDB is the same in uplink and downlink. The purpose of the PDB is to support the configuration of scheduling and link layer functions (e.g. the setting of scheduling priority weights and HARQ target operating points). The PDB shall be interpreted as a maximum delay with a confidence level of 98 percent.
> NOTE 1:	The PDB denotes a "soft upper bound" in the sense that an "expired" packet, e.g. a link layer SDU that has exceeded the PDB, does not need to be discarded (e.g. by RLC in E-UTRAN). The discarding (dropping) of packets is expected to be controlled by a queue management function, e.g. based on pre-configured dropping thresholds.

So in the 3GPP definition:
- Uplink and downlink PDB are the same value
- The PDB is not a maximum allowed latency, but the value above which the packet could be dropped (by schedulers), and will be considered dropped (for packet loss calculations). So for a specified packet error loss ratio of 1%, up to 1% of packets are allowed to exceed the specified PDB.

But fixed networks may require different values for uplink and downlink, so we may just need to live with 3GPP profiles needing to specify the same value for both.

Note that the latency between PCEF and the IGW will be minimal, so my preference would be to define PDB as the upper bound on latency between UE and IGW, which is the genuine "edge" of a CSP network.

jlurien: I agree that the term "packetDelayBudget" is more accurate for what we can offer. It can be left to the description to explain that it refers to the one-way latency between the device and the gateway from the operator's network to other networks, and that the E2E latency will be about two times this value plus the latency not controlled by the operator. 
The most relevant attributes to differentiate profiles are the assigned priority and rate intervals, while other parameters such as this one are rather derivative and informative.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,236 @@ components:
...
+          example: 2.2
+          description: |
+            Priority levels allow efficient resource allocation and ensure optimal performance
+            for various services in each technology, with the highest priority traffic receiving
+            preferential treatment.
+            The lower value the higher priority.
+            Since not all access networks use the same priority range, so this priority will be
+            scaled to the access network's priority range.
+          format: float
+          minimum: 1
```
eric-murray, line 498: When reviewing 3GPP QCI / 5QI definitions, I noted that:
- For 5QI, they multiplied the QCI priorities by 10, so priorities are now an integer < 100
- For QCI, there are priorities defined that are higher than (so numerically lower than) 1, specifically QCI 65 (0.7) and QCI 69 (0.5)

I'd suggest that `priority` is either:
- an integer between 1 and 99; or
- a float between 0 and 10

My preference is probably the latter, but either would work for 3GPP networks.

hdamker:
@RandyLevensalor Would it be possible to resolve the comments from @eric-murray or do you need more input on them?

RandyLevensalor:
@eric-murray @jlurien Thanks for the additional feedback.  Please let me know if the latest change addressed your comments.

@hdamker I believe that everything to date has been resolved.  🤞

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,229 @@ components:
...
+        maxUpstreamRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+              The maximum best effort data
+        maxUpstreamBurstRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+              The burst rate for best effort data
```
jlurien, line 448: I think it is not clear what is this in relation to `maxUpstreamRate`. If a provider designs a service to be "statistically" limit between `targetMinUpstreamRate` and  `maxUpstreamRate`, what does the burst rate represent.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,229 @@ components:
...
+            - description: |
+                The name of the QoS profile.  This name is used to identify the profile when creating a session.
+                The name must be unique.
+        status:
+          allOf:
+            - $ref: "#/components/schemas/QosProfileStatusEnum"
+            - description: |
+              The status of the QoS profile.  The status is set to ACTIVE when the profile is created.
+              The status is set to INACTIVE when the profile is deleted.
+        targetMinUpstreamRate:
```
jlurien, line 421: I miss a new optional property:
```
QoSProfile:
  properties:
    description:
      description: Qualitative description of the profile
      type: string
```    
This is a demand from our Product team to leave some placeholder where to put indications or disclaimers not easily covered in other properties

jlurien (commented):
Thanks @RandyLevensalor,

please consider adding a new property `description` to the `QoSProfile`, and some clarification to the burst rate properties

code/API_definitions/qod-api.yaml
RandyLevensalor: What mix / max length would you suggest for the description property?

Should we call it qosDescription or something that's not an openid keyword?

code/API_definitions/qod-api.yaml
RandyLevensalor: @jlurien This is an optional parameter without a default.  So if you service don't have a burst rate then it shouldn't define one.

Do we need update all of the optional values to define the QoS profile behaviour if that parameter isn't defined?

code/API_definitions/qod-api.yaml
jlurien: To me `description`  is enough, as it is already under `qosProfile`, so context is clear. I don't think it is necessary to define limits explicitly, but I don't have strong opinion here, if any, a max limit in the order of 1000 would be enough.

code/API_definitions/qod-api.yaml
jlurien: Yes, I get that it's optional. Comment was to enhance the description of "burst rate", I feel it is not something easily understood without further explanation in this context.

eric-murray: It should at least be clarified that this is NOT the 3GPP "Maximum Burst Size", which is measured in bytes, not bits/second. But I agree a proper definition would be better. As "burst rate" is not a 3GPP concept, I guess it comes from fixed networks?

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
       description: |
         Ports may be specified as a list of ranges or single ports.
     QosProfile:
+      description: |
+        Data type with attributes of a QosProfile
+      type: object
+      properties:
+        name:
+          allOf:
```
jlurien, line 410: I've just checked with redocly viewer and this does not work well when the schema in $ref already has a description, as the composed schema will have 2 descriptions and the one in the $ref'ed schema prevails. So if 2 different descriptions are needed for a same $ref'ed schema,  the $ref'ed schema must have no description and the description is moved to the `allOf`

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+            - $ref: "#/components/schemas/QosProfileName"
+            - description: |
```
jlurien, line 412: In this case the description in `QosProfileName` is more complete, so I would remove this one.

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+                The name must be unique.
+        description:
+          description: |
+            A description of the QoS profile.
+          type: string
+          example: "QoS profile for video streaming"
+        status:
+          allOf:
+            - $ref: "#/components/schemas/QosProfileStatusEnum"
+            - description: |
```
jlurien, line 423: To keep this clarification it has to be moved to the description in `QosProfileStatusEnum`

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+                The status of the QoS profile.  The status is set to ACTIVE when the profile is created.
+                The status is set to INACTIVE when the profile is deleted.
+        targetMinUpstreamRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
```
jlurien, line 428: Remove description in `Rate` to keep the relevant one

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+            - description: |
+                When defined, this is the maximum downstream burst rate for the QoS profile will enable
+                the network to burst data at a higher rate than the maxDownstreamRate for a period of time.
+
+                This can result in improved user experience when there is additional network capacity.
+                For instance, when a user is streaming a video, the network can burst data at a higher rate
+                to fill the buffer, and then return to the maxUpstreamRate once the buffer is full.
+        minDuration:
+          allOf:
+            - $ref: "#/components/schemas/Duration"
```
jlurien, line 490: Remove description in `Duration` to keep the relevant one

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+          - A searchable descriptive name
+      type: string
+      example: QCI_1_voice
+      minLength: 3
+      maxLength: 256
+      format: string
+      pattern: "^[a-zA-Z0-9_.-]+$"
+    Rate:
+      type: object
+      description: Data rate value and unit
```
jlurien, line 570:
```suggestion
```

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+          type: integer
+          example: 10
+          format: int32
+          minimum: 0
+          maximum: 1024
+        unit:
+          $ref: "#/components/schemas/RateUnitEnum"
+    Duration:
+      type: object
+      description: Value and unit of time
```
jlurien, line 582:
```suggestion
```

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+        maxUpstreamRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+                The maximum best effort data
+        maxUpstreamBurstRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+                When defined, this is the maximum upstream burst rate for the QoS profile will enable
```
jlurien, line 453:
```suggestion
                When defined, this is the maximum upstream burst rate for the QoS profile, that will enable
```

code/API_definitions/qod-api.yaml
```diff
@@ -329,18 +402,238 @@ components:
...
+        maxDownstreamRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+                The maximum best effort rate
+        maxDownstreamBurstRate:
+          allOf:
+            - $ref: "#/components/schemas/Rate"
+            - description: |
+                When defined, this is the maximum downstream burst rate for the QoS profile will enable
```
jlurien, line 482:
```suggestion
                When defined, this is the maximum downstream burst rate for the QoS profile, that will enable
```

jlurien (changes requested):
Some adjustments needed in allOf and multiple descriptions to show the right one

hdamker:
As discussed: @sfnuser @eric-murray @jlurien please do the final check (and) if possible approval until June 6th. Feel free to create issues for later improvements of the descriptions - the goal is to get the PR merged if possible asap.

sfnuser (dismissed):
Thanks @RandyLevensalor for driving this to conclusion.

eric-murray:
@hdamker
It's difficult to review the overall API definition file because Randy's fork is some 80 commits behind the main branch. Whilst much of this is other files, there have been some changes qod-api.yaml that are not included in Randy's PR.

@RandyLevensalor
Can you sync your fork with the main branch (and fix any conflicts that may arise)? This has to be done sometime, so might as well do it now.

RandyLevensalor:
> @hdamker It's difficult to review the overall API definition file because Randy's fork is some 80 commits behind the main branch. Whilst much of this is other files, there have been some changes qod-api.yaml that are not included in Randy's PR.
> 
> @RandyLevensalor Can you sync your fork with the main branch (and fix any conflicts that may arise)? This has to be done sometime, so might as well do it now.

@eric-murray I've rebased the branch.

Surprisingly enough, there were no conflicts and I didn't see any regressions.  Though an additional set of eyes on this will make me feel better.

code/API_definitions/qod-api.yaml
```diff
@@ -347,18 +420,213 @@ components:
...
+        - Nanoseconds
+    QosProfileStatusEnum:
       description: |
-        * `QOS_E` - Qualifier for enhanced communication profile
-        * `QOS_S` - Qualifier for the requested QoS profile _S_
-        * `QOS_M` - Qualifier for the requested QoS profile _M_
-        * `QOS_L` - Qualifier for the requested QoS profile _L_
-    
+        The current status of the QoS Profile
+        - Active - QoS Profile is available to be used
```
jlurien, line 614:
```suggestion
        - `ACTIVE`- QoS Profile is available to be used
```

code/API_definitions/qod-api.yaml
```diff
@@ -347,18 +420,213 @@ components:
...
+        - Inactive - QoS Profile is not currently available to be deployed
```
jlurien, line 615:
```suggestion
        - `INACTIVE`- QoS Profile is not currently available to be deployed
```

code/API_definitions/qod-api.yaml
```diff
@@ -347,18 +420,213 @@ components:
...
+        - Deprecated - QoS profile is actively being used in a QoD session, but can not be deployed in new QoD sessions
```
jlurien, line 616:
```suggestion
        - `DEPRECATED`- QoS profile is actively being used in a QoD session, but can not be deployed in new QoD sessions
```

code/API_definitions/qod-api.yaml
```diff
@@ -216,6 +216,79 @@ paths:
...
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/QosProfilesNotFound404"
+        "500":
+          $ref: "#/components/responses/QoSProfile500"
+        "503":
+          $ref: "#/components/responses/Generic503"
+  /qos-profiles/{qos-profile-name}:
```
jlurien, line 259:
```suggestion
  /qos-profiles/{name}:
```

Better as the name of the property, with the description, the context is undestood

code/API_definitions/qod-api.yaml
```diff
@@ -216,6 +216,79 @@ paths:
...
+    get:
+      tags:
+        - QoS Profiles
+      summary: "Get QoS Profile for a given name"
+      operationId: getQosProfile
+      description: |
+        Returns a QoS Profile that matches the given name.
+      parameters:
+        - name: qos-profile-name
```
jlurien, line 268:
```suggestion
        - name: name
```

jlurien (approved):
Some final suggestions and I give the approval in advance. We can refine parameters and descriptions in subsequent iterations

RandyLevensalor:
@jlurien Thanks for the additional suggestions.  I've merged them into the PR.

jlurien (dismissed):

code/API_definitions/qod-api.yaml
```diff
@@ -347,18 +420,213 @@ components:
...
+      type: string
+      enum:
+        - ACTIVE
+        - INACTIVE
+        - DEPRECATED
+    RateUnitEnum:
+      type: string
+      enum:
+        - bps
+        - Kbps
```
eric-murray, line 626: This should be `kbps`. Symbol for "kilo" is "k". See [SI Prefixes](https://en.wikipedia.org/wiki/Metric_prefix).

eric-murray (dismissed):

hdamker:
LGTM - thanks @sfnuser @jlurien @eric-murray for the detailed reviews and your approvals. I will try to delete the patchfile.diff and merge.

hdamker (approved):

PR #139, opened 2023-04-20
Updated method for identifying devices by IPv4 address

### LINKED ISSUE #34: Resolving ueAddr ambiguity when it is the UE private address

eric-murray:
#### What type of PR is this?

* enhancement/feature
* documentation

This PR introduces the new schema `DeviceIpv4Addr`, which is used for the `ipv4Address` property of the `Device` object. The `DeviceIpv4Addr` object allows the `device` to be identified by a combination of its public IPv4 address (`publicAddress`) and at least one of either its public port (`publicPort`) or private (device local) IPv4 address (`privateAddress`).

This modified schema is required to support the scenario where it is desired to identify the `device` by its private (local) IPv4 address, as this alone is generally insufficient to identify the device, and the public (NATed) IPv4 address must also be provided.

Fixes #34 : **Resolving ueAddr ambiguity when it is the UE private address**

- For NAT64 scenarios, it has been confirmed that the device IPv6 address alone is sufficient to identify it, and hence this PR does not affect those scenarios
- Flow rules are unaffected by this PR, with the exception that IPv4 subnets can no longer be specified for the device. If `publicPort` is specified to identify the `device`, this value is **not** used for flow rules. Description of `devicePorts` has been added to clarify that these are the local ports used directly by the device for flow rules.

#### Changelog input

```
Updated qod-api.yaml
- Schema objects DeviceIpv4Addr and SingleIpv4Addr defined
- Device property ipv4Address now of type DeviceIpv4Addr
- Updated error message definitions
- Minor format changes

Updated documentation QoD_API.md
```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -288,8 +288,9 @@ components:
           $ref: "#/components/schemas/Device"
         applicationServer:
           $ref: "#/components/schemas/ApplicationServer"
-        deviceAllocatedPorts:
+        devicePorts:
           $ref: "#/components/schemas/PortsSpec"
+          description: The ports used locally by the device for flows to which the requested QoS profile should apply. If omitted, then the qosProfile will apply to all flows between the device and the specified application server address and ports.
         applicationServerPorts:
           $ref: "#/components/schemas/PortsSpec"
         qosProfile:
@@ -471,7 +472,6 @@ components:
 
     SingleIpv4Addr:
       type: string
-      format: ipv4
       description: A single IPv4 address with no subnet mask
       example: "84.125.93.10"
       pattern: '^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$'
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -68,13 +68,13 @@ paths:
                     status: 400
                     code: INVALID_ARGUMENT
                     message: "Expected property is missing: device"
-                InsufficientProperties:
+                InsufficientDeviceProperties:
                   summary: Device must be identified by at least one parameter
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
                     message: "Insufficient properties specified: device"
-                InconsistentProperties:
+                InconsistentDeviceProperties:
                   summary: Device parameters provided identify different devices
                   value:
                     status: 400
@@ -442,7 +442,7 @@ components:
       type: string
       pattern: '^\+?[0-9]{5,15}$'
       example: "123456789"
-      description: Subscriber number (MSISDN) in E.164 format, starting with country code and ptionally prefixed with '+'.
+      description: Subscriber number (MSISDN) in E.164 format, starting with country code and optionally prefixed with '+'.
 
     DeviceIpv4Addr:
       type: object
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -104,12 +104,12 @@ paths:
                     status: 400
                     code: INVALID_ARGUMENT
                     message: "Expected property is missing: qosProfile"
-                InvalidDeviceAllocatedPortsRanges:
-                  summary: Invalid port ranges specified for deviceAllocatedPorts 
+                InvalidDevicePortsRanges:
+                  summary: Invalid port ranges specified for devicePorts 
                   value:
                     status: 400
                     code: OUT_OF_RANGE
-                    message: "Invalid port ranges specified: deviceAllocatedPorts"
+                    message: "Invalid port ranges specified: devicePorts"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -81,11 +81,11 @@ paths:
                     code: INVALID_ARGUMENT
                     message: "Expected property is missing: device.ipv4Address or device.ipv6Address"
                 DevicePortsRequired:
-                  summary: devicePorts is required
+                  summary: devicePublicPorts is required
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: devicePorts"
+                    message: "Expected property is missing: devicePublicPorts"
                 QoSProfileRequired:
                   summary: qosProfile is required
                   value:
@@ -97,7 +97,7 @@ paths:
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Ranges not allowed: devicePorts"
+                    message: "Ranges not allowed: devicePublicPorts"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
@@ -223,6 +223,7 @@ components:
       type: string
       format: uuid
       description: Session ID in UUID format
+
     SessionInfo:
       description: Session related information.
       allOf:
@@ -253,6 +254,7 @@ components:
               type: array
               items:
                 $ref: "#/components/schemas/Message"
+
     CreateSession:
       description: Data type with attributes required for creating a session
       type: object
@@ -274,7 +276,7 @@ components:
           $ref: "#/components/schemas/Device"
         applicationServer:
           $ref: "#/components/schemas/ApplicationServer"
-        devicePorts:
+        deviceAllocatedPorts:
           $ref: "#/components/schemas/PortsSpec"
         applicationServerPorts:
           $ref: "#/components/schemas/PortsSpec"
@@ -293,10 +295,12 @@ components:
         - device
         - applicationServer
         - qosProfile
+
     Port:
       type: integer
       minimum: 0
       maximum: 65535
+
     PortsSpec:
       type: object
       minProperties: 1
@@ -328,6 +332,7 @@ components:
           - 5070
       description: |
         Ports may be specified as a list of ranges or single ports.
+
     QosProfile:
       type: string
       enum:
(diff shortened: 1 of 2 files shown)


code/API_definitions/qod-api.yaml
```diff
@@ -414,14 +432,50 @@ components:
...
+    DeviceIpv4Addr:
+      type: object
+      properties:
+        publicAddress:
+          $ref: "#/components/schemas/SingleIpv4Addr"
+        privateAddress:
+          $ref: "#/components/schemas/SingleIpv4Addr"
+        publicPort:
+          $ref: "#/components/schemas/Port"
+      oneOf:
```
jlurien, line 455: This `oneOf` triggered some debate. To me is acceptable at this level as it may be used by JSON-schema validators. 
An alternative `anyOf` would allow to send all 3 properties, which is something that I'm not sure if want to prevent.
I have to further validate this also with our dev team.

code/API_definitions/qod-api.yaml
```diff
@@ -274,7 +288,7 @@ components:
           $ref: "#/components/schemas/Device"
         applicationServer:
           $ref: "#/components/schemas/ApplicationServer"
-        devicePorts:
+        deviceAllocatedPorts:
```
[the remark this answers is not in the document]
eric-murray: We can include alternative names for this field in the vote being organised. Both your suggestions are reasonable. I will include a description as well, which may be ignored by machines, but can certainly be read by humans.

code/API_definitions/qod-api.yaml
```diff
@@ -414,14 +432,50 @@ components:
...
+      oneOf:
+        - required: [publicAddress, privateAddress]
+        - required: [publicAddress, publicPort]
+      example:
+            {
+              "publicAddress": "84.125.93.10",
+              "publicPort" : 59765
+            }
+      description: |
+        The device should be identified by either the public (observered) IP address and port as seen by the application server, or the private (local) and any public (observed) IP addresses in use by the device (this information can be obtained by various means, for example from some DNS servers).
```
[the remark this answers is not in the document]
eric-murray: Thanks. One day SmartBear will include a spell checker in the Swagger editor!

code/API_definitions/qod-api.yaml
eric-murray: My view was that this was the best way of presenting valid options to the API consumer. It does not prevent them sending all 3 parameters if known, but does not mandate it either, as both `privateAddress` and `publicPort` could be difficult to obtain for different scenarios.

For implementation, discriminators continue to cause problems, and may do for some time. My view if that implementors should be prepared to modify the CAMARA API definition for our own tooling. A certain amount of that has to be done anyway, for `apiRoot` or `securitySchemes` for example. In Vodafone, for example, we need to change all the lowerCamelCase `operationId`s to snake_case, but that is fine as it does not affect the API consumer. Workarounds for discriminators are also possible if not natively supported by tooling.

emil-cheung:
@hdamker ,
please help to arrange a community voting for:

Observed IP address/Local IP address vs. Public IP address/Private IP address

eric-murray:
@emil-cheung 
I had put some suggestions in the issue, but we can refine them here. Do you still want `assigned` to be considered as was proposed [here](https://github.com/camaraproject/QualityOnDemand/issues/34#issuecomment-1471407018)? If not, then I think we have the following options:

### Vote 1: `DeviceIpv4Addr` object properties
**Option 1 :** `publicAddress`, `privateAddress` & `publicPort`
**Option 2:** `observedAddress`, `localAddress` & `observedPort`


### Vote 2: Rename of `devicePorts` property of `CreateSession` object 
**Option 1 :** `deviceLocalPorts`
**Option 2 :** `deviceSourcePorts`

emil-cheung:
> @emil-cheung I had put some suggestions in the issue, but we can refine them here. Do you still want `assigned` to be considered as was proposed [here](https://github.com/camaraproject/QualityOnDemand/issues/34#issuecomment-1471407018)? If not, then I think we have the following options:
> 
> ### Vote 1: `DeviceIpv4Addr` object properties
> **Option 1 :** `publicAddress`, `privateAddress` & `publicPort` **Option 2:** `observedAddress`, `localAddress` & `observedPort`
> 
> ### Vote 2: Rename of `devicePorts` property of `CreateSession` object
> **Option 1 :** `deviceLocalPorts` **Option 2 :** `deviceSourcePorts`

Yes, we prefer 'local' instead of 'assigned'.

code/API_definitions/qod-api.yaml
eric-murray: @jlurien 

If I understand correctly, then:
```
{
  "publicAddress": "84.125.93.10",
  "privateAddress": "10.10.10.10",
  "publicPort" : 59765
}
```
would be rejected because it matches both:
```
    DeviceIpv4Addr:
      type: object
      properties:
        publicAddress:
          $ref: "#/components/schemas/SingleIpv4Addr"
        privateAddress:
          $ref: "#/components/schemas/SingleIpv4Addr"
        publicPort:
          $ref: "#/components/schemas/Port"
      required: [publicAddress, privateAddress]
```
and
```
    DeviceIpv4Addr:
      type: object
      properties:
        publicAddress:
          $ref: "#/components/schemas/SingleIpv4Addr"
        privateAddress:
          $ref: "#/components/schemas/SingleIpv4Addr"
        publicPort:
          $ref: "#/components/schemas/Port"
      required: [publicAddress, publicPort]
```
I can see that might be true (I have no way to test these things), and can see `anyOf` is "more" unambiguous (to machines at least, if not to developers), so I'll change `oneOf` to `anyOf`.

The reason to allow all three parameters is that it is much easier to identify a device from `publicAddress` and `privateAddress` than from `publicAddress` and `publicPort`. So if the API caller knows `publicAddress`, we should encourage them to give it, even if they also know `publicPort`.

jlurien: Yes, you understood right. Some guidelines recommend to use `anyOf` in most cases instead of oneOf, as it is more efficient for validators. oneOf requires to validate each schema and assure that only one is true, while validation of anyOf can return true when first validation succeeded

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
eric-murray: Decision is to leave this as `devicePorts` for now

eric-murray:
Ready for final review now

code/API_definitions/qod-api.yaml
```diff
@@ -414,14 +433,49 @@ components:
...
     NetworkAccessIdentifier:
       type: string
       example: "123456789@domain.com"
+
     PhoneNumber:
       type: string
       pattern: '^\+?[0-9]{5,15}$'
       example: "123456789"
-      description: Subscriber number in E.164 format (starting with country code). Optionally prefixed with '+'.
+      description: Subscriber number (MSISDN) in E.164 format, starting with country code and ptionally prefixed with '+'.
```
akoshunyadi, line 445: typo: ptionally

code/API_definitions/qod-api.yaml
```diff
@@ -57,47 +57,59 @@ paths:
...
+                DeviceMissing:
+                  summary: Device must be specified
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: device.phoneNumber"
-                Ipv4AddressRequired:
-                  summary: ipv4Address is required as part of device
+                    message: "Expected property is missing: device"
+                InsufficientProperties:
```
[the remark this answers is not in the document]
eric-murray: Fixed

code/API_definitions/qod-api.yaml
```diff
@@ -57,47 +57,59 @@ paths:
...
+                  summary: Device must be identified by at least one parameter
                   value:
                     status: 400
                     code: INVALID_ARGUMENT
-                    message: "Expected property is missing: device.ipv4Address"
-                IpAddressRequired:
-                  summary: Some IP address is required as part of device
+                    message: "Insufficient properties specified: device"
+                InconsistentProperties:
```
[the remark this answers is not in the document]
eric-murray: Fixed

code/API_definitions/qod-api.yaml
eric-murray: Fixed

sfnuser (commented):
Looks good to me.

sfnuser (approved):

jlurien (approved):

jlurien:
@hdamker I see that there will be a a lot of conflicts between this PR and the pending one I have to incorporate and update doc within spec, so I will wait to have this merged. If done in the next days I will submit mine before next meeting.

hdamker (approved):

hdamker:
Seems that we are good to go ... 3+ final reviews and 4 days no new comments.

Issue #144, opened 2023-05-05
Add documentation into API definition for release v0.9.0


hdamker:
**Problem description**

Documentation is not in sync with the API definition (some PRs changed the API documentation in parallel, some not).

**Expected action**

Add documentation to API definition as discussed in https://github.com/camaraproject/WorkingGroups/issues/164

**Additional context**

Agreement is that CAMARA documentation should be embedded within the OAS definition, and not separate. This reduces the effort required in maintaining documentation.

ShutingQing:
@jlurien hi jose, have u started working on this? if not, i can help on this.

PR #149, opened 2023-05-10
Modify create session url name and add a complimentary description for qos in api definition 

ShutingQing:
#### What type of PR is this?
* documentation

From developer perspective, when  i look into the api definitions, i may have some suggestions.

1. For the create session url,  currently it's '/sessions' why not use ‘createSessions’ directly？How do you guys feel?
2. For the QoS profile, through it's been clarified very clearly in documentations, but i'll suggest a brief description of qos_e qos_l qos_m qos_s is required. So people can easily know. 
3. Delete useless .gitignore file.

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/MAINTAINERS.MD
+++ b/MAINTAINERS.MD
@@ -1,11 +1,12 @@
-| Org                    | Name                                                |
-| -----------------------| ----------------------------------------------------|
-| Deutsche Telekom AG | Herbert Damker |
-| Ericsson | Emil Zhang |
-| Orange | Sylvain Morel |
-| Telefonica | Jose Luis Urien |
-| Telefonica | Jesus Peña| 
-| Vodafone | Eric Murray |
-| Verizon/ 5GFF| Syed Rehman |
-| Verizon/ 5GFF| Mahesh Chapalamadugu |
-| KDDI | Toshi (Toshiyasu) Wakayama |
+| Org                 | Name                       |
+|---------------------|----------------------------|
+| Deutsche Telekom AG | Herbert Damker             |
+| Ericsson            | Emil Zhang                 |
+| Huawei              | Shuting Qing               |
+| Orange              | Sylvain Morel              |
+| Telefonica          | Jose Luis Urien            |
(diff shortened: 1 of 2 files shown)


hdamker:
@ShutingQing Please follow the best practice to split different topics into different issues/PRs, at least into single commits. 

Here we have a mix of completely different topics. Also we have the policy to first open an issue, discuss the proposal within the issue and only then open a PR based on the support there (see https://github.com/camaraproject/Governance/blob/main/ProjectStructureAndRoles.md)

In particular:

> * For the create session url,  currently it's '/sessions' why not use ‘createSessions’ directly？How do you guys feel?

That would be a major change and would require first a discussion within an issue. But your proposal would be also a violation of the Design Guidelines related to [URL definitions](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#41-url-definition), e.g. "verbs are not allowed".

> * For the QoS profile, through it's been clarified very clearly in documentations, but i'll suggest a brief description of qos_e qos_l qos_m qos_s is required. So people can easily know.

That would fit into the PR @jlurien is working on for #144 ... but we don't want to have concrete values in the YAML file. 

> * Delete useless .gitignore file.
ok, I will do it as simple housekeeping.

In addition you have added you name to the maintainer file as part of the PR. There is no need to be in the maintainer file for the first contributions. See https://github.com/camaraproject/Governance/blob/main/ProjectStructureAndRoles.md for the definition of maintainers.

I propose to close this PR.

ShutingQing:
Hi @hdamker, thanks for suggestions. I agree it’s good to split different topics, so it will be easier for people to review. I’ll look into your references. 

My bad. Before when we contributed in onap, maybe in my project, we didn’t put too much attention on this. Cuz usually the code is a lot. So when we do a code review, we will just go through each line, and leave comments on each line we have questions. And then contributors will make changes again based on review changes till a consensus.

PR #162, opened 2023-06-09
Add PATCH operation

### LINKED ISSUE #47: Include support for /PATCH sessions
Currently the API spec does not have support for making any updates once the qos session is created.
Rather than assuming the data passed during session creation will always be static for the duration of the session, we need to consider use cases where either the UE details or the Application server details could change due to various factors and provide support for developers to make the updates via a PUT /sessions API.

emil-cheung:
#### What type of PR is this?

* enhancement/feature

* Update an existing QoD session

Fixes #47 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -160,6 +160,42 @@ paths:
           $ref: "#/components/responses/Generic500"
         "503":
           $ref: "#/components/responses/Generic503"
+    patch:
+      tags:
+        - QoS sessions
+      summary: "Patch session information"
+      operationId: patchSession
+      parameters:
+        - name: sessionId
+          in: path
+          description: Session ID that was obtained from the patchSession operation
+          required: true
+          schema:
+            $ref: "#/components/schemas/SessionId"
+      requestBody:
+        description: Patch an existing session
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/PatchSession"
+        required: true
+      responses:
+        "200":
+          description: Contains information about active session
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SessionInfo"
+        "401":
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/SessionNotFound404"
+        "500":
+          $ref: "#/components/responses/Generic500"
+        "503":
+          $ref: "#/components/responses/Generic503"
     delete:
       tags:
         - QoS sessions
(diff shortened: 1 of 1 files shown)


eric-murray:
@emil-cheung
I didn't think the discussion on what parameters could be patched had been concluded, so my questions on your proposal are:
- Why allow `applicationServer` to be updated? Whilst this may be technically possible, the new server might not be compatible with the agreed QoS profile or business / legal considerations (e.g. prioritising traffic the new application server might not be compliant with net neutrality rules).

  And what happens if the PATCH request is refused because the new application server is not allowed? Does traffic continue to be prioritised to the previous application server, even though that may not be required.

  My preference would be to not allow `applicationServer` to be patched, and instead require a new session to be created if a change in application server is required.

- Why can `qosProfile` not be patched? I would have thought that a more likely scenario than changing application server. For example, the end user application might find that the QoS profile currently being used does not give the desired QoE, and wants to upgrade without changing other parameters.

- Does the patched `duration` still apply from the original session creation time? So sessions are still limited to one day, and this cannot be extended by patching?

- The `notificationUrl` and `notificationAuthToken` format does not match that being proposed in #155, which is following the updated CAMARA design guidelines. Also, updating `notificationAuthToken` is understandable, by why would `notificationUrl` be updated during a session?

hdamker:
On `duration` we need to decide the semantics:
* does the duration values still counts from begin of the session? What happens then if PATCH sets a duration which is already exceeded by the session?
* or will the duration be counted from the (successful) PATCH operation onwards? (only that will allow to achieve sessions longer than the defined maximum of 86400 seconds)

hdamker:
A general comment: I consider this PR as one for the release after v0.9.0. It will need intensive discussions before it can be merged and would therefore postpone the release otherwise.

emil-cheung:
> A general comment: I consider this PR as one for the release after v0.9.0. It will need intensive discussions before it can be merged and would therefore postpone the release otherwise.

@hdamker 
It is for v0.10.0 and no hurry.
As the v0.9.0 comes to the end, I provide a draft proposal for the community to further discuss.

jlurien:
Agree that we have to discuss which parameters it makes sense to allow to be updated. Requirement should come from what it is convenient and intuitive for the client, not just because it is technically feasible to do it in the network.

emil-cheung:
> Agree that we have to discuss which parameters it makes sense to allow to be updated. Requirement should come from what it is convenient and intuitive for the client, not just because it is technically feasible to do it in the network.

@jlurien @eric-murray @hdamker 
Sorry for moving to quickly. 
Let's park the PR at this moment and go back to the #47 to discuss the changeable attributes.
I have created a table to scan the attributes in Session information.

emil-cheung:
> On `duration` we need to decide the semantics:
> 
> * does the duration values still counts from begin of the session? What happens then if PATCH sets a duration which is already exceeded by the session?
> * or will the duration be counted from the (successful) PATCH operation onwards? (only that will allow to achieve sessions longer than the defined maximum of 86400 seconds)

@hdamker 
My thought is the timer will be reset since the successful PATCH operation, which is simple and clear.

hdamker:
@emil-cheung: i changed the PR to "draft" status, until we have discussed the behavior behind patching `duration` withing in the issue #47 and PR is adapted based on the result.

(all people who have subscribed to the PR previously are still subscribed)

PR #177, opened 2023-06-30
fix(openapi): Add inheritance beetween Event and QosStatusChangedEvent

### LINKED ISSUE #174: Simplify Event Model for Notification in v0.9.0
I'm not sure to understand why we have Event, EventBase and QosStatusChangedEvent. 
IMHO we can define an Event that contains all EventBase properties and a specialized version QosStatusChangedEvent.
See below.
   
``` yaml
 Event:
      description: The event being notified
      type: object
      required:
        - eventId
        - eventType
        - eventTime
      properties:
        eventId:
          $ref: "#/components/schemas/EventId"
        eventType:
          $ref: "#/components/schemas/EventType"
        eventTime:
          $ref: "#/components/schemas/EventTime"
      discriminator:
        propertyName: eventType
        mapping:
          QOS_STATUS_CHANGED: "#/components/schemas/QosStatusChangedEvent"

QosStatusChangedEvent:
      type: object
      allOf:
          - $ref: "#/components/schemas/Event"
      required:
          - sessionId
          - qosStatus
      properties:
          sessionId:
                $ref: "#/components/schemas/SessionId"
          qosStatus:
                $ref: "#/components/schemas/EventQosStatus"
          statusInfo:
                $ref: "#/components/schemas/StatusInfo"



```
This way, returning a QosStatusChangedEvent as an event is quite simple because QosStatusChangedEvent is derived from Event. There's also no need to group specific properties into an eventDetail, someone who wants to ignore the detailed information will just have to cast the specialized object to an Event.

patrice-conil:
#### What type of PR is this?
* enhancement/feature

Fix ability to send QosStatusChangedEvent from client side (Telco) to the server side of /notifications (QoS API Consumer) for those who use strongly typed Languages.
As QosStatusChangedEvent extends Event we can now send it as payload

Fixes #174

~~BREAKING CHANGE: Remove EventNotification to embed Event as notification payload~~ (reverted, see [bbe5812](https://github.com/camaraproject/QualityOnDemand/pull/177/commits/bbe5812db50f1b6bd0fa3f6b58bade2213a1fd70))

This specification replaces multiple inheritance schemes with single inheritance. This way the generator will stop to generate anonymous XXXAllOf.java classes which are not useful at all.

```yaml
ClassA:
   allOf:
      - $ref: ClassB
      - type: object
            properties:
                 prop1:
 ```   
That says ClassA inherit from ClassB and anonymous inlined object containing property prop1 is replaced by:

```yaml
ClassA:
   type: object
   allOf:
      - $ref: ClassB
   properties:
      prop1:
 ```   
That means: ClassA extends ClassB and contains property prop1.

#### Changelog input

```
Add inheritance beetween Event and QosStatusChangedEvent
Simplify Notification payload model and align it with DeviceStatus proposal 

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -742,7 +742,17 @@ components:
         eventSubscriptionId:
           type: string
           description: 'The ID of explicit subscription or none in case of implicit subscription'
-          example: '7698d710-98b7-4695-a958-7b228f08128c'
+      example:
+        {
+          "eventId": "5698d710-9b1b-4695-a958-7b228f08128c",
+          "eventType": "QOS_STATUS_CHANGED",
+          "eventTime": "2023-05-30T10:18:28Z",
+          "eventDetail": {
+            "sessionId": "7698d710-98b7-4695-a958-7b228f08128c",
+            "qosStatus": "UNAVAILABLE",
+            "statusInfo": "DURATION_EXPIRED"
+          }
+        }
 
     Event:
       description: The event being notified
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -749,6 +749,17 @@ components:
         propertyName: eventType
         mapping:
           QOS_STATUS_CHANGED: "#/components/schemas/QosStatusChangedEvent"
+      example:
+        {
+          "eventId": "5698d710-9b1b-4695-a958-7b228f08128c",
+          "eventType": "QOS_STATUS_CHANGED",
+          "eventTime": "2023-05-30T10:18:28Z",
+          "eventDetail": {
+            "sessionId": "7698d710-98b7-4695-a958-7b228f08128c",
+            "qosStatus": "UNAVAILABLE",
+            "statusInfo": "DURATION_EXPIRED"
+          }
+        }
 
     EventId:
       type: string
@@ -797,7 +808,7 @@ components:
           type: string
           enum:
             - DURATION_EXPIRED
-            - NETWORK_TERMINATED¡
+            - NETWORK_TERMINATED
 
     Device:
       description: |
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -739,7 +739,7 @@ components:
       properties:
         event:
           $ref: "#/components/schemas/Event"
-        subscriptionId:
+        eventSubscriptionId:
           type: string
           description: 'The ID of explicit subscription or none in case of implicit subscription'
           example: '7698d710-98b7-4695-a958-7b228f08128c'
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -113,7 +113,7 @@ paths:
                 content:
                   application/json:
                     schema:
-                      $ref: "#/components/schemas/EventNotification"
+                      $ref: "#/components/schemas/Event"
               responses:
                 "204":
                   description: Successful notification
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jlurien** [comment]
This new proposal to model inheritance may work but I have concerns about not being totally aligned with the [OAS specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#schemaComposition), which states that:

> The OpenAPI Specification allows combining and extending model definitions using the `allOf` property of JSON Schema, in effect offering model composition. `allOf` takes an array of object definitions that are validated independently but **together compose a single object.**

You are now moving the item that defined the child event out of the array. Even if this JSON-schema is valid and may work for code generation, it is not what it is shown in all examples in the specs, which follow the approach to model the child properties within the `allOf`

https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

>
```
    ExtendedErrorModel:
      allOf:     # Combines the BasicErrorModel and the inline model
        - $ref: '#/components/schemas/BasicErrorModel'
        - type: object
          required:
            - rootCause
          properties:
            rootCause:
              type: string
```

I think it is safer to follow the examples in the spec as much as possible, as others may claim that their tools don't work well with this simplification.

**Review by jlurien — CHANGES_REQUESTED** [review]
As commented in https://github.com/camaraproject/WorkingGroups/pull/255#pullrequestreview-1510504032, It is OK to model the event following inheritance instead of polymorphism but we should stick as much as possible to the examples in the spec

**Comment by patrice-conil** [comment]
Comes back to common usage of allOf following suggestions of @jlurien

**Review by jlurien — APPROVED** [review]
LGTM

**Review by hdamker — DISMISSED** [review]
/LGTM

**Comment by hdamker** [comment]
@akoshunyadi @eric-murray can you have a final look before I merge?

**Comment by eric-murray** [comment]
Small point, but the Swagger editor does not create a valid example for the notification request body, so I'd suggest to put an example of type `QosStatusChangedEvent` within the definition of the `Event` schema to give a better idea as to what a notification will look like.

But otherwise it looks fine and I'll add my approval.

**Review by akoshunyadi — APPROVED** [review]
LGTM, but then the guideline should be updated, which contains that additional layer of eventNotification

**Comment by patrice-conil** [comment]
@eric-murray, 
QOS_STATUS_CHANGED added as Event example

**Review by eric-murray — APPROVED** [review]

**Comment by hdamker** [comment]
An important point we should clarify before we merge:

> BREAKING CHANGE: Remove EventNotification to embed Event as notification payload

The guideline defines `eventNotification` with `event` as one of it's attributes. The current PR  would require an update of the Design Guidelines, as @akoshunyadi rightfully commented. As far as I can see DeviceLocation is following the schema of the guideline in https://github.com/camaraproject/DeviceStatus/pull/31

@patrice-conil What's the rational for this change?
@bigludo7 as author of the guideline, could you review our PR as well and comment?

**Comment by patrice-conil** [comment]
> An important point we should clarify before we merge:
> 
> > BREAKING CHANGE: Remove EventNotification to embed Event as notification payload
> 
> The guideline defines `eventNotification` with `event` as one of it's attributes. The current PR would require an update of the Design Guidelines, as @akoshunyadi rightfully commented. As far as I can see DeviceLocation is following the schema of the guideline in [camaraproject/DeviceStatus#31](https://github.com/camaraproject/DeviceStatus/pull/31)
> 
> @patrice-conil What's the rational for this change? @bigludo7 as author of the guideline, could you review our PR as well and comment?

@hdamker,
As a follower of the [KISS](https://dev.to/kwereutosu/the-k-i-s-s-principle-in-programming-1jfg) principle, I just try to keep it as simple as possible. As EventNotification does not provide any additional data and requires the creation of an additional object before posting the notification I've proposed to remove it in the PR.
I agree it opens a new discussion... but as it will be reused a lot and is quite structuring maybe it's worth it (in commonalities?).
What is the original motivation for defining an EventNotification that simply encapsulates an Event in a field? Do we plan to notify anything else? 
I will be on vacation this Friday, If you can save 5 min to discuss this point with our partners,  I will follow the decision of the working group and adapt the model.
Sorry to open so many questions but it is with use that we realize the problems.

BR,
Patrice

**Comment by jlurien** [comment]
I think the reason to have an `event` encapsulating the data is because for notifications coming from **explicit** subscriptions we may have a `subscriptionId`, next to the event:

```
{
 "eventSubscriptionId": "456g899g",
 "event": {
    ...
  }
}
```

In QoD we have implicit notifcations`, and as Patrick comments it adds an additional level.

The decision should come from Commonalities. We may decide to have move all attributes of `event` to the parent level, next to `eventSubscriptionId`,  and remove the `event` object.

patrice-conil:
@hdamker, @jlurien,
I missed the eventSubscriptionId ... thanks Jose
What if to keep the commonalities we keep eventSubscriptionId in EventNotification (we could set it with sessionId)?
What do you think about?

hdamker:
@jlurien @patrice-conil My view is that if we wait for a commonality decision now, we would delay the release for several weeks. I would go with the current structure in commonalities and the proposal [above](https://github.com/camaraproject/QualityOnDemand/pull/177#issuecomment-1634308370) from @jlurien. But that's just my view. 

Beyond that I expect changes of the notification/subscription structure as soon there will be common OAS for notification endpoint on listener side(see https://github.com/camaraproject/Commonalities/issues/8). We need to adapt in upcoming release.

shilpa-padgaonkar:
Please take a look at https://github.com/camaraproject/Commonalities/issues/8#issuecomment-1629615969 and provide your feedback in commonalities

patrice-conil:
Hi @hdamker, @jlurien,
Thank you for your constructive comments,
I've reintroduced EventNotification to the spec... hope this helps moving forward soon.

hdamker (approved):
/LGTM

hdamker:
@jlurien @eric-murray @akoshunyadi @bigludo7 Just waiting for one other approval of the latest changes before I merge.

akoshunyadi (approved):

bigludo7 (approved):
LGTM

jlurien (approved):
LGTM

eric-murray:
Happy to approve to progress this, but at some point we should add an explicit example for the `EventNotification` callback, because the example generated by the Swagger editor includes a value for `eventSubscriptionId`, which is not correct.

eric-murray (approved):

patrice-conil:
> Happy to approve to progress this, but at some point we should add an explicit example for the `EventNotification` callback, because the example generated by the Swagger editor includes a value for `eventSubscriptionId`, which is not correct.

Example added thanks to @eric-murray

hdamker (approved):

PR #205, opened 2023-08-17
Move "description" out of "allOf" declaration

### LINKED ISSUE #204: Move the "description" out of "allOf"-declaration
**Problem description**
"allOf" including a $ref and a 'description' is not compatible with openapi - generators for later implemantation usage.

```yaml
    CreateSession:
      type: object
      properties:
        <...>
        devicePorts:
          allOf:
            - description: The ports used locally by the device for flows to which the requested QoS profile should apply. If omitted, then the qosProfile will apply to all flows between the device and the specified application server address and ports
            - $ref: "#/components/schemas/PortsSpec" 
        applicationServerPorts:
          allOf:
            - description: A list of single ports or port ranges on the application server
            - $ref: "#/components/schemas/PortsSpec"
        <...>
```

**Expected behavior**
The description should be out of the "allOf".

```yaml
    CreateSession:
      <...>
        devicePorts:
          description: The ports used locally by the device for flows to which the requested QoS profile should apply. If omitted, then the qosProfile will apply to all flows between the device and the specified application server address and ports
          allOf:
            - $ref: "#/components/schemas/PortsSpec"
        applicationServerPorts:
          description: A list of single ports or port ranges on the application server
          allOf:
            - $ref: "#/components/schemas/PortsSpec"
         <...>
```

maxl2287:
#### What type of PR is this?

* correction

We are currently discovering that by using an openapi generator the current version 0.9.0 of QoD causes problems by having the "description" inside of "allOf"-declaration

Fixes #204

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -451,12 +451,12 @@ components:
         applicationServer:
           $ref: "#/components/schemas/ApplicationServer"
         devicePorts:
+          description: The ports used locally by the device for flows to which the requested QoS profile should apply. If omitted, then the qosProfile will apply to all flows between the device and the specified application server address and ports
           allOf:
-            - description: The ports used locally by the device for flows to which the requested QoS profile should apply. If omitted, then the qosProfile will apply to all flows between the device and the specified application server address and ports
-            - $ref: "#/components/schemas/PortsSpec" 
+            - $ref: "#/components/schemas/PortsSpec"
         applicationServerPorts:
+          description: A list of single ports or port ranges on the application server
           allOf:
-            - description: A list of single ports or port ranges on the application server
             - $ref: "#/components/schemas/PortsSpec"
         qosProfile:
           $ref: "#/components/schemas/QosProfileName"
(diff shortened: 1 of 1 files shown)


maxl2287:
If there is a way to add it to v0.9.0 it would be quite good.

RandyLevensalor:
@maxl2287 Please see comment in the issue.

Also, many of these AllOf statements were added specifically for the description and could be removed if we were to remove the description from the AllOf.

hdamker:
> Also, many of these AllOf statements were added specifically for the description and could be removed if we were to remove the description from the AllOf.

@RandyLevensalor @maxl2287 have you tested or can you test that the behavior of editors and tooling is then still the expected one? I thought that we needed the `allof `workaround to get it done.

jlurien:
> > Also, many of these AllOf statements were added specifically for the description and could be removed if we were to remove the description from the AllOf.
> 
> @RandyLevensalor @maxl2287 have you tested or can you test that the behavior of editors and tooling is then still the expected one? I thought that we needed the `allof `workaround to get it done.

It works well with both Redoc and Swagger viewers. The aim of the "allOf" was to avoid having $ref as property at same level of description, but with the workaround, this is avoided.

eric-murray:
Hi @maxl2287

I think you need to rebase this PR, as github is reporting a conflict. Can you do that? Thanks.

maxl2287:
> Hi @maxl2287
> 
> I think you need to rebase this PR, as github is reporting a conflict. Can you do that? Thanks.

Yep, done.

eric-murray (approved):

hdamker (approved):

RandyLevensalor (approved):

PR #207, opened 2023-08-24
Megalinter initial

RandyLevensalor:
#### What type of PR is this?

Add one of the following kinds:

* tests

#### What this PR does / why we need it:

This adds megalinter github action to lint yaml files and provide some basic repository checks.

This checks against the openapi configuration with a few example rules based on commonalities.

Whitespace changes were made to the yaml files to pass the yaml linting.

A couple of minor changes were made to ensure compliance with the open api schema.

#### Which issue(s) this PR fixes:

Fixes #158 

#### Special notes for reviewers:

The yamllint and openapirc configuration files are in the root of the repository.  These could be moved to another directory.

JAVA and MARKDOWN linters, which are a part of megalinter are currently disabled.  Enabling these will require additional changes to their respective formats and possible updates to the linting.

#### Changelog input

```
 release-note

Added megalinter github action to check the formatting of yaml files.

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #215, opened 2023-09-07
Enable java linting in megalinter with linting rules

ravindrapalaskar17:
As part of this **PR: -**  https://github.com/camaraproject/QualityOnDemand/pull/207 megalinter added in QOD repository.
But JAVA linting is disabled in above PR also linting rules is not defined.  
 
This PR is for enable the JAVA linting in QOD repository:

1. Enabled the JAVA linting and megalinter.yaml
2. Set JAVA linting rule in javalinting.xml


eric-murray:
Hi @ravindrapalaskar17

Can you run these linting rules on the current QoD release (0.9.0) to show what changes it would request?

A recent update to the [API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#11-definition-in-openapi) recognised that inappropriate use of keywords in an API OAS definitions could cause problems with code generators, so I'm open to this being enforced by linting rules.

But we need to be careful not to be overly restrictive in how we construct our OAS definitions simply to avoid problems that are fundamentally those of the code generators themselves and not of OAS.

hdamker:
@eric-murray:

> Can you run these linting rules on the current QoD release (0.9.0) to show what changes it would request?

As discussed within the call: linting of YAML files is already active since #207. In #207 you can also find the  (minor) changes within qod-api.yaml to adapt to linting. Currently we have not many CAMARA specific rules defined ... there is https://github.com/camaraproject/Commonalities/issues/15 to develop them in line with the CAMARA Design Guidelines.

The current PR here will only add the linting of JAVA files (with a custom ruleset which fits to the used codestyle ... might need to revised if there will a project wide JAVA codestyle in future).

RandyLevensalor (approved):

PR #216, opened 2023-09-07
Extend session duration

### LINKED ISSUE #47: Include support for /PATCH sessions

### LINKED ISSUE #220: Change duration of an active session
**Problem description**
Currently there is no mean to change the duration of an active session. However, there are use cases require extending the duration. A dedicated endpoint is proposed for this operation.

**Possible evolution**
N/A

**Alternative solution**
Use PATCH operation to change the duration.

**Additional context**
Originally, PATCH operation was proposed to change the duration. However, according the further discussion in issue #47 , a dedicated endpoint is preferred since it is straightforward to developers. This new issue is created to reflect the community discussion.

emil-cheung:
#### What type of PR is this?

enhancement/feature

Extend the duration of an active session.

Fixes #220 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -296,7 +296,7 @@ paths:
         - QoS Sessions
       summary: "Extend the duration of an active session"
       description: |
-        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the additional duration will be the maximum limit minus the remaining duration.
+        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the new remaining duration will be the maximum limit.
       operationId: extendQosSessionDuration
       parameters:
         - name: sessionId
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -322,7 +322,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/SessionInfo"
         "400":
-          description: Invalid input for createSession operation
+          description: Invalid input for extendQosSessionDuration operation
           content:
             application/json:
               schema:
@@ -593,7 +593,7 @@ components:
           maximum: 86399
           example: 60
       required:
-        - duration
+        - extendedDuration
 
 
     QosProfile:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -297,7 +297,9 @@ paths:
       summary: "Extend the duration of an active session"
       description: |
         Extend the duration of an active QoS session.
-        If this operation is executed successfully, the new duration of the target session will be the original duration plus extended duration.
+        If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration.
+        The remaining duration plus the requested additional duration shall not exceed the maximum limit. 
+        Otherwise, the additional duration will be the maximum limit minus the remaining duration.
       operationId: extendQosSessionDuration
       parameters:
         - name: sessionId
@@ -339,12 +341,6 @@ paths:
                     status: 400
                     code: INVALID_ARGUMENT
                     message: "The target session is inactive"
-                ExceedMaxDuration:
-                  summary: The extended session duration exceeds the limit
-                  value:
-                    status: 400
-                    code: INVALID_ARGUMENT
-                    message: "The extended session duration exceeds the limit of 86,400 seconds"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
@@ -591,7 +587,7 @@ components:
       description: Attributes required to extend the duration of an active session
       type: object
       properties:
-        additionalDuration:
+        requestedAdditionalDuration:
           description: |
             Additional duration in seconds to be extended.
           type: integer
@@ -600,7 +596,7 @@ components:
           maximum: 86399
           example: 60
       required:
-        - additionalDuration
+        - requestedAdditionalDuration
 
 
     QosProfile:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -293,12 +293,11 @@ paths:
   /sessions/{sessionId}/extend:
     post:
       tags:
-        - Extend or reduce QoS session duration
-      summary: "Extend or reduce QoS session duration"
+        - Change session duration
+      summary: "Extend the duration of an active session"
       description: |
-        Extend or reduce the duration of an active QoS session.
-        If this operation is executed successfully, the new duration of the target session will be original duration plus extended duration.
-        If the new duration is shorter than or equal to elapsed duration, the session will be terminated.
+        Extend the duration of an active QoS session.
+        If this operation is executed successfully, the new duration of the target session will be the original duration plus extended duration.
       operationId: extendQosSessionDuration
       parameters:
         - name: sessionId
@@ -308,7 +307,7 @@ paths:
           schema:
             $ref: "#/components/schemas/SessionId"
       requestBody:
-        description: Parameters to extend or reduce the duration of an active session
+        description: Parameters to extend the duration of an active session
         content:
           application/json:
             schema:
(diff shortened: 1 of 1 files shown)


code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,67 @@ paths:
...
+  /sessions/{sessionId}/extend:
+    post:
+      tags:
+        - Extend or reduce QoS session duration
+      summary: "Extend or reduce QoS session duration"
+      description: |
+        Extend or reduce the duration of an active QoS session.
+        If this operation is executed successfully, the new duration of the target session will be original duration plus extended duration.
+        If the new duration is shorter than or equal to elapsed duration, the session will be terminated.
+      operationId: extendQosSessionDuration
```
patrice-conil, line 302: As you can use operation to increase/decrease duration I think we can name operation changeQosSessionDuration.

emil-cheung: Well, I do have a theoretical use case in mind but I would like to validate with the community.

Let's say a teleoperator uses a QoD session to boost the connection for remotely controlling a car. 
As the initial ETA (Estimated Time of Arrival) of the trip is one hour, the teleoperating application server creates a session with duration 60 minutes. Along the trip, since the traffic situation is better than expected, the ETA is adjusted to 45 minutes, the teleoperating application server may reduce the session duration to 45 minutes.

hdamker: > Well, I do have a theoretical use case in mind but I would like to validate it with the community.
> 
> Let's say a teleoperator uses a QoD session to boost the connection for remotely controlling a car. As the initial ETA (Estimated Time of Arrival) of the trip is one hour, the teleoperating application server creates a session with duration 60 minutes. Along the trip, since the traffic situation is better than expected, the ETA is adjusted to 45 minutes, the teleoperating application server may reduce the session duration to 45 minutes.

@emil-cheung well, I would consider this use case also theoretically :-) In the use case it is much more pratical to `DELETE` the session explicitly when the car arrived at the destination and/or no longer needs the quality boost.But the other way round - extending the session duration if the journey takes longer then expected - is for sure an important use case.

My proposal is to limit the scope of the issue and the PR to extending the session duration and omit "reduce" from it (I have seen that the issue is already only mentioning extension).

eric-murray: I am also in favour of only allowing extensions to the session duration, and not reductions.

An additional point. If session duration reduction is allowed, then a reduction that would reduce the session duration to zero or below should be considered an error, rather than just deleting the session. There are two reasons for this:
- it probably is an error, because if the API consumer had actually wanted to delete the session, they could do that explicitly
- the required scope for this endpoint would probably be `qod:sessions:update` according to [this](https://github.com/camaraproject/Commonalities/pull/57) PR in Commonalties, and we would then be allowing a `delete` with only `update` scope

emil-cheung: As several companies prefer starting with only duration extension, I have updated both the issue and the PR to duration extension only. If in the future, there is need for duration reduction, let's open a new issue to improve.

code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,66 @@ paths:
...
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/SessionNotFound404"
+        "500":
+          $ref: "#/components/responses/Generic500"
+        "503":
+          $ref: "#/components/responses/Generic503"
+
```
jlurien, line 352:
```suggestion

      security:
        - oAuth2ClientCredentials: []
        - threeLegged:
            - "qod-sessions-write"
```

The scope for write is shared with the operation to create session. We could differentiate between -create and -update but -write covers both.

code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,66 @@ paths:
...
+                  value:
+                    status: 400
+                    code: INVALID_ARGUMENT
+                    message: "Schema validation failed at  ..."
+                InactiveSession:
+                  summary: The target session is inactive
+                  value:
+                    status: 400
+                    code: INVALID_ARGUMENT
+                    message: "The target session is inactive"
```
jlurien, line 341: It would be useful to include an example of an error when the required extension is over the maximum session duration. This is something that we should also make clear in the description, if remaining + extension <= max session duration.

jlurien (commented):
Agree with extension only. 2 comments:
- security is missing
- we should make clear behaviour and errors regarding remaining + extension <= maximum duration

code/API_definitions/qod-api.yaml
jlurien: Are we allowing remaining + extension > maximum session duration. If not we should add this a constraint in the description and consider an error for that.

code/API_definitions/qod-api.yaml
emil-cheung: It is added.

code/API_definitions/qod-api.yaml
emil-cheung: It is added.

code/API_definitions/qod-api.yaml
RandyLevensalor: @jlurien has a good point.  This should not be a workaround for the maximum session duration.

Would we even further constrain this so that initial duration + extension < max session duration?

hdamker: Within last QoD call we have discussed to constrain in such way, that the resulting new remaining session duration should not extend the maximum session duration (of 86400 seconds).

So: 
IF remaining + extension <= maximum session duration, duration will be set to remaining duration + extension

We also discussed the case if remaining duration + extension > maximum session duration in the extension request. In this case there was the notion that we should not raise an error, but set the session duration to the value where the new remaining duration is equal to the maximum session duration (as the developer can't calculate the accurate maximum extension and there would be several race conditions possible).

The all cases the actual new duration will be in the response.

code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,69 @@ paths:
...
+            schema:
+              $ref: "#/components/schemas/ExtendSessionDuration"
+        required: true
+      responses:
+        "200":
+          description: Contains information about active session
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SessionInfo"
```
eric-murray, line 321: Returning full session info here seems excessive.

Why not return only the fields that have changed (`duration`, `expiresAt`) plus `sessionId` itself?

code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,69 @@ paths:
         - threeLegged:
             - "qod-sessions-delete"
 
+  /sessions/{sessionId}/extend:
+    post:
+      tags:
+        - Change session duration
```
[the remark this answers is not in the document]
jlurien: I wouldn't add a new tag for this. Tags are intended to group operations and this operation can be part of the existing "QoS Sessions" tag.

code/API_definitions/qod-api.yaml
```diff
@@ -290,6 +290,69 @@ paths:
...
+      summary: "Extend the duration of an active session"
+      description: |
+        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the additional duration will be the maximum limit minus the remaining duration.
```
jlurien, line 299:
```suggestion
        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the new remaining duration will be the maximum limit.
```

code/API_definitions/qod-api.yaml
jlurien: I would keep just the extension, not reduction, so operation name is valid

code/API_definitions/qod-api.yaml
jlurien: I would see acceptable to return the new modified SessionInfo, no strong opinion on this

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
hdamker: Addressed in https://github.com/camaraproject/QualityOnDemand/pull/216/commits/867c74384f8dec5c7cbcf1120342e54b388a72ee

code/API_definitions/qod-api.yaml
hdamker: Agreed within meeting on Nov 16th to return full SessionInfo.

hdamker (approved):
LGTM

hdamker:
@eric-murray @RandyLevensalor Please have a final view

RandyLevensalor (approved):

hdamker:
No further comments since last week, so we are good here.

PR #217, opened 2023-09-08
issue_194_PR

### LINKED ISSUE #194: Add Application Function Id (afId) or Sponsor Id
**Problem description**
 Service Provider would like to track QoD session data usage for third party usage reporting or monetization purposes. 

**Possible evolution**
 Add Sponsor Id or Application Function ID (afID) in QoD API. These IDs are defined in 3GPP specs TS 29.122 (SCEF)  and TS 29.522 (NEF).  This will enable QoD session usage tracking based on application function ID in the network.

SyeddR:
#### What type of PR is this?
* PR for introducing application consumer Id as discussed in #194 

* enhancement/feature

* Identifies the application consumer that initiates a qos session. The ID is systematically relayed to downstream systems, including but not limited to, the Network Exposure Function (NEF) and the Policy Control Function (PCF) within the 5G framework. The primary utility of this transmission is to facilitate the application of relevant policies, ensuring that the application consumer is appropriately tagged within usage records.
     

Fixes #194 

For lack of the better term i am proposing "applicationConsumerId" as string type field. Open for suggestions

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -433,12 +433,7 @@ components:
       description: Attributes required to create a session
       type: object
       properties:
-        applicationConsumerId:
-          description: |
-            Identifies the application consumer that initiates a qos session. The ID is systematically relayed to downstream systems, including but not limited to, the Network Exposure Function (NEF) and the Policy Control Function (PCF) within the 5G framework.
-            The primary utility of this transmission is to facilitate the application of relevant policies, ensuring that the application consumer is appropriately tagged within usage records.
-          type: string
-          example: XRapp123
+
         duration:
           description: |
             Session duration in seconds. Maximal value of 24 hours is used if not set.
@@ -872,6 +867,12 @@ components:
         * `ipv6Address` 
       type: object
       properties:
+        applicationId:
+          description: |
+            A public identifier that uniquely identifies the application initiating a QoS session. The applicationId is more granular than the clientId, which identifies the unique customer. A client or customer could have one or more applications that need to be tracked. The Id is systematically relayed to downstream systems, including but not limited to, the Network Exposure Function (NEF) and the Policy Control Function (PCF) within the 5G framework.
+            The primary utility of this transmission is to facilitate the application of relevant policies, ensuring that the application is appropriately tagged within usage records.
+          type: string
+          example: XRapp123
         ipv4Address:
           $ref: "#/components/schemas/Ipv4Address"
         ipv6Address:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -433,7 +433,6 @@ components:
       description: Attributes required to create a session
       type: object
       properties:
-
         duration:
           description: |
             Session duration in seconds. Maximal value of 24 hours is used if not set.
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -858,12 +858,14 @@ components:
 
     ApplicationServer:
       description: |
-        A server hosting backend applications to deliver some business logic to clients.
+        A server hosting backend applications to deliver a service to the identified `device`.
         
-        The developer can choose to provide the below specified device identifiers:
-
+        The developer must provide at least one of the following device identifiers:
         * `ipv4Address`
-        * `ipv6Address` 
+        * `ipv6Address`
+
+        In addition, the developer can specify an `applicationId` for policy control / audit purposes. The `applicationId` is not sufficient in itself to identity the application server.
+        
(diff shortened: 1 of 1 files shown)


eric-murray:
Hi @SyeddR

I had understood that the `applicationConsumerId` must be associated with one, and only one application. Is that the case? If so, my preference would be for this to be a sub-property of the `applicationServer` property, the reason being that I'm not a fan of "flat" input parameter structures, and much prefer that related parameters be grouped, particularly when some (such as this one) are optional.

Of course, you could argue that this should also apply to `applicationServerPorts`, and I'd agree with you :-)

I also agree that adding this property to the `ApplicationServer` schema requires some thought, as we need to ensure that the API consumer provides at least one IP address, but there are ways in OAS that that can be mandated whilst still allowing `applicationConsumerId` to be optionally specified.

SyeddR:
> Hi @SyeddR
> 
> I had understood that the `applicationConsumerId` must be associated with one, and only one application. Is that the case? If so, my preference would be for this to be a sub-property of the `applicationServer` property, the reason being that I'm not a fan of "flat" input parameter structures, and much prefer that related parameters be grouped, particularly when some (such as this one) are optional.
> 
> Of course, you could argue that this should also apply to `applicationServerPorts`, and I'd agree with you :-)
> 
> I also agree that adding this property to the `ApplicationServer` schema requires some thought, as we need to ensure that the API consumer provides at least one IP address, but there are ways in OAS that that can be mandated whilst still allowing `applicationConsumerId` to be optionally specified.

jlurien:
I suggested to open an issue on Identity and Consent WG about this topic, as identification of Application is being discussed there in the context of authentication, and access token should contain all required information.

code/API_definitions/qod-api.yaml
```diff
@@ -433,6 +433,7 @@ components:
       description: Attributes required to create a session
       type: object
       properties:
+
```
eric-murray, line 436: Why the blank line here?

SyeddR: Fixed it.

eric-murray:
@SyeddR 
I'm fine with the proposal, but updated the `ApplicationServer` schema description to clarify that `applicationId` itself is not sufficient to identify the application server, and that an IP address is still required.

I'm a bit concerned that the schema definition will not catch API consumers who only provide the `applicatonId` (because that will satisfy the `minProperties: 1` directive), so am wondering if we need to re-organise the schema so that a single `ipAddress` property is now `required`. One for discussion.

There appear to be some conflicts now with the main branch (which maybe I introduced - apologies if that is the case), Can you synchronise your fork with main to see if that fixes things?

SyeddR:
> There appear to be some conflicts now with the main branch (which maybe I introduced - apologies if that is the case), Can you synchronise your fork with main to see if that fixes things?

@eric-murray My fork shows all synced with latest updates. Can you try to resolve the conflict at your end.

eric-murray (approved):

eric-murray:
@SyeddR
OK, fixed and now approved from my side. It seems my original edits were not properly merged into your own fork, but subsequent edits were, so all good now.

code/API_definitions/qod-api.yaml
```diff
@@ -858,14 +858,21 @@ components:
...
         * `ipv6Address`
+
+        In addition, the developer can specify an `applicationId` for policy control / audit purposes. The `applicationId` is not sufficient in itself to identity the application server.
       type: object
       properties:
+        applicationId:
+          description: |
+            A public identifier that uniquely identifies the application initiating a QoS session. The applicationId is more granular than the clientId, which identifies the unique customer. A client or customer could have one or more applications that need to be tracked. The Id is systematically relayed to downstream systems, including but not limited to, the Network Exposure Function (NEF) and the Policy Control Function (PCF) within the 5G framework.
+            The primary utility of this transmission is to facilitate the application of relevant policies, ensuring that the application is appropriately tagged within usage records.
+          type: string
```
RandyLevensalor, line 874: Could we some constraints on the string size and values?
minLength: 3
maxLength: 256 (default value)
pattern: (should we exclude strings like "??? or ***"

RandyLevensalor (commented):
Could we apply some constraints to this?

hdamker:
Converted to draft status as agreed within our call on Nov 3rd.

code/API_definitions/qod-api.yaml
RandyLevensalor: @SyeddR any response to this?

hdamker:
This PR is stale and can't be merged into the current (splitted) API structure anymore.

PR #227, opened 2023-10-17
Adding global tags element 

### LINKED ISSUE #223: Operation tags in global tags element
**Problem description**

Spectral linting rule [operation-tag-defined](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#operation-tag-defined)
produces warnings for qod-api.yaml like this:

```
  Message :   Operation tags must be defined in global tags.
  Rule    :   operation-tag-defined
  Path    :   paths./sessions.post.tags.0
  Line    :   88
```

**Expected behavior**
Tags should be defined in a top-level `tags` element

**Alternative solutions**
1. Removing tags from operation objects in qod-api.yaml 

2. Exclusion of this rule from future ruleset for CAMARA APIs as `tags` element is **not required** by OAS

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#fixed-fields

> A list of tags used by the specification with additional metadata. The order of the tags can be used to reflect on their order by the parsing tools. Not all tags that are used by the [Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#operationObject) must be declared. The tags that are not declared MAY be organized randomly or based on the tools' logic. Each tag name in the list MUST be unique.

**Additional context**
 Spectral default OAS ruleset includes also:

[openapi-tags-alphabetical](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#openapi-tags-alphabetical)

> 
> OpenAPI object should have alphabetical tags. This will be sorted by the name property.
> Recommended: No

[tag-description](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#tag-description)

> Tags alone are not very descriptive. Give folks a bit more information to work with.
> ```
> 
> tags:
>   - name: "Aardvark"
>     description: Funny-nosed pig-head raccoon.
>   - name: "Badger"
>     description: Angry short-legged omnivores.
> ```
> 
> If your tags are business objects then you can use the term to explain them a bit. An 'Account' could be a user account, company information, bank account, potential sales lead, or anything. What is clear to the folks writing the document is probably not as clear to others.
> Recommended: No

rartych:
#### What type of PR is this?

* correction

Operation tags should be defined in a top-level `tags` element.

Fixes #223

Description field is rendered by Swagger this way:
QoD_tags_in_Swagger

#### Changelog input

```
Global tags element added

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -82,7 +82,7 @@ servers:
         default: qod/v0
         description: Base path for the QoD API
 tags:
-  - name: QoS sessions
+  - name: QoS Sessions
     description: Manage QoS sessions
   - name: QoS Profiles
     description: Manage QoS Profiles
@@ -90,7 +90,7 @@ paths:
   /sessions:
     post:
       tags:
-        - QoS sessions
+        - QoS Sessions
       summary: Creates a new session
       description: Create QoS Session to manage latency/throughput priorities
       operationId: createSession
@@ -231,7 +231,7 @@ paths:
   /sessions/{sessionId}:
     get:
       tags:
-        - QoS sessions
+        - QoS Sessions
       summary: Get QoS session information
       description: Querying for QoS session resource information details
       operationId: getSession
@@ -268,7 +268,7 @@ paths:
 
     delete:
       tags:
-        - QoS sessions
+        - QoS Sessions
       summary: Delete a QoS session
       description: Free resources related to QoS session
       operationId: deleteSession
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -85,7 +85,7 @@ tags:
   - name: QoS sessions
     description: Manage QoS sessions
   - name: QoS Profiles
-    description: Manage QoS Profiles   
+    description: Manage QoS Profiles
 paths:
   /sessions:
     post:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -81,6 +81,11 @@ servers:
       basePath:
         default: qod/v0
         description: Base path for the QoD API
+tags:
+  - name: QoS sessions
+    description: Manage QoS sessions
+  - name: QoS Profiles
+    description: Manage QoS Profiles   
 paths:
   /sessions:
     post:


code/API_definitions/qod-api.yaml
```diff
@@ -81,6 +81,11 @@ servers:
       basePath:
         default: qod/v0
         description: Base path for the QoD API
+tags:
+  - name: QoS sessions
+    description: Manage QoS sessions
+  - name: QoS Profiles
+    description: Manage QoS Profiles
```
hdamker, line 88:
```suggestion
    description: Manage QoS Profiles
```

hdamker (commented):
One small correction necessary to make linting happy ;-)

jlurien (commented):
We should use same format for both tags, either both sessions and profiles in lowecase or uppercase. Tags per operation have to be adjusted accordingly

hdamker:
@jlurien

> We should use same format for both tags, either both sessions and profiles in lowecase or uppercase. Tags per operation have to be adjusted accordingly

I looked over the documentation parts, and there we have also the same difference, it's more or less consequently "QoS session" and "QoS Profile". But I propose to not extend the scope of this PR to a cleanup of the documentation parts, and do it here only for the tags. 

I would tend to lowercase "QoS session" and "QoS profile", but that's just my view. Any other views?

eric-murray:
> I would tend to lowercase "QoS session" and "QoS profile", but that's just my view. Any other views?

My preference is Title Case, but this is a style point and really one for Commonalities and the design guidelines., whose [example tags](https://github.com/camaraproject/Commonalities/blob/main/documentation/images/guidelines-fig-15.png) are unfortunately ambiguous on this point.

rartych:
I do no think we need exact guidelines for tags style/format, as tags are helper properties to clearly describe and render OAS and are useful with large API specifications.
I have aligned to the current style of tags, with the reasoning that while sessions are dynamic and created on request, QoD Profiles were initially predefined and static.
But I agree that having just "QoS session" and "QoS profile" is good solution.
Should the tags and their descriptions be updated in this PR?
Or together with separate issue for unification of the style in all the documentation?

eric-murray:
If we want a common "look and feel" between CAMARA APIs, then we need an agreement in Commonalities on tag format style.

RandyLevensalor (approved):
LGTM  Thanks for adding this.

eric-murray (approved):

hdamker (approved):

hdamker:
Good to go. Thanks @rartych !

PR #228, opened 2023-10-19
Add GET /sessions endpoint for retrieving all active sessions of authenticated user

### LINKED ISSUE #101: List endpoint for active sessions of authenticated user
# Summary

Create a `/get` endpoint offering a paginated response of all sessions that the currently authenticated user created and are still active.

# Context

While testing the APIs it might happen that the `session id` from a successful create request will get lost. In that case, there is no way to retrieve the created sessions: The `/get` endpoint only returns sessions whose ids are known, but not all active ones for the authenticated user.

# Proposal

Add a variant of the `/get` endpoint that takes optionally page information to return all sessions for the authenticated user. We optionally need pagination information, since more sessions might be available then we can return (think response size limits etc). Therefore this endpoint must be paginated.

The default values for the paginated parameters *current page* would be set to 0 and the *size of a page* to 100.  So on average no page information is needed, but for heavy users, it is available.

The sorting of the response could be by it's `startedAt` field: sessions created early should appear before later sessions.

## cURL example

This is an example of the proposed endpoint:

A call like

```
curl --request GET '{{baseUrl}}/sessions?page=4&pageSize=10' --header 'Authorization: <token>'
```

would return

```
{
  "pageSize": 10,
  "pageNumber": 4,
  "sessions": [
    {
      "id": "f0a49ce4-bda0-76eb-879c-3f19db88e985",
      "qos": "QOS_M",
      "ueId": {
         /// ....
      },
      "asId": {
         /// ....
      },
      "duration": 10
      "expiresAt": 1639566000,
      "startedAt": 1639479600,
    },
    /// 9 more sessions
  ]
}
```

# Alternatives

A) One of the main reason for this proposal is the inability to find created sessions. This mostly stems from testing and due to user error the session id cannot be retrieved from the initial create request. In order to not block the ue and only create the needed amount of sessions, the default of the `duration` parameter could be set to a very low value: a 10 seconds duration would be deleting the session after 10 seconds, without the need to delete a session without the session id.

B) A second alternative could be to add a *delete all* endpoint: With this endpoint, we would not give out the session information, but could guarantee in an error case, there is the ability to delete existing sessions.

# Initial Specification Proposal

Under #100 you can find an initial draft of a proposal on how the specification would need to change. (closed now, till a decisions on this issue is found)

kacper-kicinski:
#### What type of PR is this?
* feature

* This PR adds a new GET /sessions endpoint for retrieving all active sessions of the authenticated user

Fixes #101

image

#### Changelog input

```
GET /sessions endpoint for retrieving all active sessions of authenticated user added.
```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -253,13 +253,28 @@ paths:
               schema:
                 $ref: "#/components/schemas/SessionInfoList"
         "400":
-          $ref: "#/components/responses/Generic400"
+          description: Invalid pagination for getSession operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+              examples:
+                PageOutOfRange:
+                  summary: The page number is out of the range
+                  value:
+                    status: 400
+                    code: INVALID_PAGINATION
+                    message: "Invalid pagination: page number out of the range"
+                InvalidPerPageValue:
+                  summary: The perPage value is less than 1
+                  value:
+                    status: 400
+                    code: INVALID_PAGINATION
+                    message: "Invalid pagination: perPage value can't be less than 1"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
           $ref: "#/components/responses/Generic403"
-        "404":
-          $ref: "#/components/responses/Generic404"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -229,11 +229,11 @@ paths:
       description: Get information about all active sessions authorized to be retrieved by the provided access token.
       operationId: getSessions
       parameters:
-      - name: phoneNumber
-        in: query
-        required: false
-        description: Allows filtering of sessions by device (phone number).
-        schema:
+        - name: phoneNumber
+          in: query
+          required: false
+          description: Allows filtering of sessions by device (phone number).
+          schema:
             $ref: "#/components/schemas/PhoneNumber"
       responses:
         "200":
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -226,19 +226,8 @@ paths:
       tags:
         - QoS sessions
       summary: Get all active sessions information
-      description: Get information about all active sessions that were created by the client
+      description: Get information about all active sessions authorized to be retrieved by the provided access token.
       operationId: getSessions
-      parameters:
-        - name: perPage
-          in: query
-          required: false
-          schema:
-            $ref: "#/components/schemas/PerPage"
-        - name: page
-          in: query
-          required: false
-          schema:
-            $ref: "#/components/schemas/Page"
       responses:
         "200":
           description: Contains information about the complete list of active sessions
@@ -246,31 +235,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SessionInfoList"
-        "206":
-          description: Contains information about the incomplete list of active sessions
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/SessionInfoList"
-        "400":
-          description: Invalid pagination for getSession operation
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/ErrorInfo"
-              examples:
-                PageOutOfRange:
-                  summary: The page number is out of the range
-                  value:
-                    status: 400
-                    code: INVALID_PAGINATION
-                    message: "Invalid pagination: page number out of the range"
-                InvalidPerPageValue:
-                  summary: The perPage value is less than 1
-                  value:
-                    status: 400
-                    code: INVALID_PAGINATION
-                    message: "Invalid pagination: perPage value can't be less than 1"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
@@ -494,18 +458,15 @@ components:
             - qosStatus
 
     SessionInfoList:
-      description: Paged response of active sessions created by the user
+      description: List response of active sessions authorized to be retrieved by the provided access token.
       type: object
       required:
         - sessions
       properties:
-        perPage:
-          $ref: "#/components/schemas/PerPage"
-        page:
-          $ref: "#/components/schemas/Page"
         sessions:
           description: Array of sessions to be returned.
           type: array
+          minItems: 0
           items:
             type: object
             $ref: "#/components/schemas/SessionInfo"
@@ -1069,18 +1030,6 @@ components:
         - status
         - code
         - message
-
-    PerPage:
-      description: Number of resources requested to be provided in the response.
-      type: integer
-      format: int32
-      default: 20
-
-    Page:
-      description: Requested page number to indicate the start of the resources to be provided in the response (considering perPage page size). The first page has the number "0".
-      type: integer
-      format: int32
-      default: 0
 
   responses:
     Generic400:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -225,7 +225,7 @@ paths:
     get:
       tags:
         - QoS sessions
-      summary: Get all active sessions information
+      summary: Get all active session IDs
       description: Get information about all active sessions authorized to be retrieved by the provided access token.
       operationId: getSessions
       parameters:
@@ -237,11 +237,11 @@ paths:
           $ref: "#/components/schemas/PhoneNumber"
       responses:
         "200":
-          description: Contains information about the complete list of active sessions
+          description: Contains information about the complete list of active session IDs
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/SessionInfoList"
+                $ref: "#/components/schemas/SessionIdList"
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
@@ -464,18 +464,18 @@ components:
             - expiresAt
             - qosStatus
 
-    SessionInfoList:
-      description: List response of active sessions authorized to be retrieved by the provided access token.
+    SessionIdList:
+      description: List response of active session IDs authorized to be retrieved by the provided access token.
       type: object
       required:
-        - sessions
+        - sessionIds
       properties:
-        sessions:
-          description: Array of sessions to be returned.
+        sessionIds:
+          description: Array of session IDs to be returned.
           type: array
           minItems: 0
           items:
-            $ref: "#/components/schemas/SessionInfo"
+            $ref: "#/components/schemas/SessionId"
 
     CreateSession:
       description: Attributes required to create a session
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -225,7 +225,7 @@ paths:
     get:
       tags:
         - QoS sessions
-      summary: Get all active session IDs
+      summary: Get all active sessions information
       description: Get information about all active sessions authorized to be retrieved by the provided access token.
       operationId: getSessions
       parameters:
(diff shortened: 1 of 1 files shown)


RandyLevensalor:
@kacper-kicinski The linters failed.  Please check the results above and address. Thanks!

kacper-kicinski:
> @kacper-kicinski The linters failed. Please check the results above and address. Thanks!

There is some issue that would be good to discuss here:

When a client requests for the list of active sessions, but there is no one, what HTTP code and response should be returned? We opt for returning just 200 and sessions = empty array. Another solution would be returning the 404 - Not Found, but we don't recommend it since it is not really an error, just the sessions list is empty.
 
But there is also another point: since we support pagination, the client could, in the request, specify a page value which is greater than the last available page (for example, there are pages 0,1,2 and the client requests for page 3, which does not exist). In that case, 404 could be returned. But here we go to the cause of linting failure - there is no Generic404 specification, which was referred here: https://github.com/kacper-kicinski/QualityOnDemand/blob/8f394bfa592e6aca23b9e952f959dbd842590d73/code/API_definitions/qod-api.yaml#L262

Possible solutions:
1. Simply add Generic404 specification
2. Change $ref from Generic404 to SessionNotFound404, which is already specified, but - does it really fit that case?
3. Add PageNotFound404 specification and refer to it instead of Generic404
4.  Return in this case another response code, not 404 (but which one?)

For me, both 1 and 3 solutions could be applied. What do you think about it?

rartych:
Summarizing our discussion during the call:

-  When there is no active session: return 200 and sessions = empty array
- According to [Design guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#81-pagination) for `page` out of range the code 400 should be returned:
`400: request outside the range of the resource list `
It can cover:
   -   Invalid pagination parameters (page or per_page less than 1)
   -   Invalid page number (exceeding the total number of pages)

I think instead of 
```
    Generic400:
      description: Invalid input
```
new response should be defined and used here:
```
    InvalidPagination400:
      description: Invalid pagination
```
What do you think?

kacper-kicinski:
> Summarizing our discussion during the call:
> 
> * When there is no active session: return 200 and sessions = empty array
> * According to [Design guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#81-pagination) for `page` out of range the code 400 should be returned:
>   `400: request outside the range of the resource list `
>   It can cover:
>   
>   * Invalid pagination parameters (page or per_page less than 1)
>   * Invalid page number (exceeding the total number of pages)
> 
> I think instead of
> 
> ```
>     Generic400:
>       description: Invalid input
> ```
> 
> new response should be defined and used here:
> 
> ```
>     InvalidPagination400:
>       description: Invalid pagination
> ```
> 
> What do you think?

Thanks for the comment! Yes, I agree with the 400 code for invalid pagination. 

Instead of creating a new error response template (like InvalidPagination400), I simply added two examples of the 400 response: PageOutOfRange and InvalidPerPageValue, directly in the GET /sessions specification because the pagination is only supported by this endpoint: https://github.com/camaraproject/QualityOnDemand/blob/3f9f2fbefcf8b3b3a2ff0b7c04fb6127cad63ea6/code/API_definitions/qod-api.yaml#L255.

eric-murray:
Please see my [comment](https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-1772943978) in the associated issue

akoshunyadi:
Returning only the sessionIds would make filters even more useful. E.g. for MSISDN

code/API_definitions/qod-api.yaml
```diff
@@ -433,6 +493,23 @@ components:
...
+        - sessions
+      properties:
+        perPage:
+          $ref: "#/components/schemas/PerPage"
+        page:
+          $ref: "#/components/schemas/Page"
+        sessions:
+          description: Array of sessions to be returned.
+          type: array
+          items:
```
jlurien, line 509:
```suggestion
          minItems: 0
          items:
```

code/API_definitions/qod-api.yaml
```diff
@@ -433,6 +493,23 @@ components:
             - expiresAt
             - qosStatus
 
+    SessionInfoList:
+      description: Paged response of active sessions created by the user
```
jlurien, line 497: Here we mention created by the user, and above created by the client. This topic is very relevant for this endpoint as discussed in #101. The logic must be to return the sessions that are allowed to be discovered by the access token. In the most restrictive case, for three-legged tokens, that includes checking both client_id and user. 
Returning just sessionIds or all session details does not change the picture, if knowing sessionIds is enough to get session details in a separate operation.

jlurien (commented):
The topic of sessions ownership and requirements for implementations to check access tokens is now critical with the new endpoint, but it is probably something that we may need for previous  GET /session/{sessionId} as well. We assumed sessionId was a "secret", but it makes sense to validate also that the sessionId was created by the client_id and for for the user linked to the access token.

code/API_definitions/qod-api.yaml
```diff
@@ -222,7 +222,67 @@ paths:
         - oAuth2ClientCredentials: []
         - threeLegged:
             - "qod-sessions-write"
-
+    get:
+      tags:
+        - QoS sessions
+      summary: Get all active sessions information
+      description: Get information about all active sessions that were created by the client
```
jlurien, line 229: Here, we should indicate that the returned sessions will be those allowed to be discovered with the provided access token, which will depend on the type of access token provided, 2-legged or 3-legged. This has to be clarified in the implementation guidelines and aligned with Identity&Consent Management guidelines, but to progress with the spec we could add something like:

```suggestion
      description: Get information about all active sessions authorized to be retrieved by the provided access token.
```

code/API_definitions/qod-api.yaml
```diff
@@ -222,7 +222,67 @@ paths:
...
+      operationId: getSessions
+      parameters:
```
jlurien, line 231: It could be interesting to add additional filters, specially to allow filtering sessions for some specific device

code/API_definitions/qod-api.yaml
jlurien, line 497:
```suggestion
      description: Paged response of active sessions authorized to be retrieved by the provided access token.
```

code/API_definitions/qod-api.yaml
```diff
@@ -222,7 +222,67 @@ paths:
...
+        - name: perPage
```
jlurien, line 232: Paging could be interesting for 2-legged scenarios where some client_id may potentially create lots of sessions. For 3-legged scenarios, paging is probably not that necessary, as each user will likely have a limited number of active sessions. Until we fully clarify which kind of tokens will be allowed for QoD, I would not add paging parameters yet, we could add them in a subsequent review if they are needed.

kacper-kicinski:
Thank you for the comments. I removed pagination, changed some descriptions, allowed filtering of sessions by phoneNumber. For now I left the response as a list of whole sessions. However, we can continue to discuss wheter it would be better to return only sessionsIds.

code/API_definitions/qod-api.yaml
```diff
@@ -222,7 +222,38 @@ paths:
...
+      summary: Get all active sessions information
+      description: Get information about all active sessions authorized to be retrieved by the provided access token.
+      operationId: getSessions
+      parameters:
+      - name: phoneNumber
+        in: query
+        required: false
+        description: Allows filtering of sessions by device (phone number).
+        schema:
+          $ref: "#/components/schemas/PhoneNumber"
```
RandyLevensalor, line 237: Add two more spaces to resolve linting issue.

hdamker:
@kacper-kicinski Would you mind to declare the PR "ready for review"? 

We discussed in today's QoD meeting that the PR will it make only into the v0.10.0 release candidate if it can be finished (= including final reviews and approvals) together with the other three open PRs within the next week (until Friday Nov 24th). The goal is to create the release candidate at the end of next week. If the PR will be not ready to be merged until then it will not make it into v0.10.0.

code/API_definitions/qod-api.yaml
```diff
@@ -222,7 +222,38 @@ paths:
...
             - "qod-sessions-write"
-
+    get:
+      tags:
+        - QoS sessions
+      summary: Get all active sessions information
+      description: Get information about all active sessions authorized to be retrieved by the provided access token.
+      operationId: getSessions
+      parameters:
+      - name: phoneNumber
```
RandyLevensalor, line 232: It would be better to filter on $ref: "#/components/schemas/Device and not just phone number.

RandyLevensalor:
> Returning only the sessionIds would make filters even more useful. E.g. for MSISDN

Only returning sessionIds could be a much faster and more responsive API call than returning the complete session object.  Especially with filtering, where the sessionId may be sufficient.  Plus, there is an existing API call to get the session information based on sessionId.

hdamker:
Please see my [comment](https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-1818409415) within the issue where I summarized the open discussions based on our QoD call on Friday.

From that it is yet to be decided for v0.10.0
* Paging or not
* Filtering on phoneNumber only vs filtering on `device` object (comment above)

kacper-kicinski:
I changed the response from a list of whole session objects to a list of only session IDs. Please read Herbert's [comment](https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-1818409415). There are some open points for further discussion, pagination, security, filtering etc. MR is now ready for review. Kindly ask reviewers to take a look.

jlurien:
> I changed the response from a list of whole session objects to a list of only session IDs. Please read Herbert's [comment](https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-1818409415). There are some open points for further discussion, pagination, security, filtering etc. MR is now ready for review. Kindly ask reviewers to take a look.

This is not aligned with the decision we took during last [meeting](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/MeetingMinutes/MOM-2023-11-17.md):

> Should the operation deliver the full session information or just the session ids?
Result of the discussion: return the full session information
Rational: this is the more developer friendly option, as otherwise the API consumer would need to call getSession one by one to find the information they are looking for. From privacy perspective has the API consumer provided and seen all personal information already during the respective session creations. It should be enough that the API consumer was authorized to create the sessionId(s) to assume consent for the get operation.

hdamker:
Setting the PR again to "draft" as the discussions within #101 are not yet concluded and still active.

eric-murray:
Issue #101 is now active again, and I think @patrice-conil and myself have landed on a proposal that we both can live with. Can others review and comment?

The main difference is that the new endpoint will only return session details that were created by the API consumer for a given end user device. The end user device will be identified from the OAuth token, so will not need to be explicitly specified in the API definition.

For those that don't understand how this will work, see [here](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md).

jlurien:
> Issue #101 is now active again, and I think @patrice-conil and myself have landed on a proposal that we both can live with. Can others review and comment?
> 
> The main difference is that the new endpoint will only return session details that were created by the API consumer for a given end user device. The end user device will be identified from the OAuth token, so will not need to be explicitly specified in the API definition.
> 
> For those that don't understand how this will work, see [here](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md).

I have replied as well in the other issue thread, and I have drafted a proposal to change this operation to a POST method, that would ease many problems that we face with the GET. Regarding whether the device has to be provided explicitly or may be identified by the token, I think that this question is not specific to the new operation. For createSession, we should have the same requirements regarding security, in order to check that the access token allows the client to create session for certain device, and in some cases the device could be known from the access token directly, so we could make device optional for createSession as well.

hdamker:
This stale PR is superseded by #325.

PR #236, opened 2023-11-19
Creation of release candicate v0.10.0-rc

hdamker:
#### What type of PR is this?

Add one of the following kinds:

* documentation
* subproject management

#### What this PR does / why we need it:

The merge of this PR will be the target commit for the first release candidate of v0.10.0, getting the tag `v0.10.0-rc`

#### Which issue(s) this PR fixes:

Fixes #235 

#### Special notes for reviewers:

Currently draft PR, as the final information about the PRs included in the release candidate are not yet available.

#### Changelog input

NA

PR #237, opened 2023-11-20
Single IP addresses in Device model specified with standard formats instead of patterns

### LINKED ISSUE #230: Single IPv6 addresses within device
**Problem description**

When redesigning `device`, for IPv4 addresses it was considered just a single IP address, and subnets were only kept for `applicationServer`. But for IPv6 addresses we are keeping the same schema in both device and applicationServer, with same pattern that allows generic masks for subnets, without a clear differentiation.

**Expected behavior**
Behaviour should be consistent for Ipv4 and Ipv6 and only allow IPv6 addresses in device which identify a single device.

We may need to differentiate schemas with dedicated descriptions and maybe patterns. Defining an exact pattern for Ipv6 is complex, so an option would be to remove them from spec and just rely on the RFC 5952 reference. 

As this model is reused by other APIs in other WGs, they may be impacted as well.

jlurien:
#### What type of PR is this?

* enhancement/feature

* IPv6 address in device changed to SingleIpv6Adress. IPv6 address in ApplicationServer is not changed
* Patterns for single IPv4 and IPv6 adresses, used in device model, substituted for the standards ipv4 and ipv6 formats defined in JSON-schema

Fixes #230 

#### Changelog input

```
* Single IP addresses in Device model specified with standard formats instead of patterns

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -858,7 +858,7 @@ components:
         ipv4Address:
           $ref: "#/components/schemas/DeviceIpv4Addr"
         ipv6Address:
-          $ref: "#/components/schemas/Ipv6Address"
+          $ref: "#/components/schemas/SingleIpv6Address"
       minProperties: 1
 
     ApplicationServer:
@@ -917,8 +917,14 @@ components:
     SingleIpv4Addr:
       description: A single IPv4 address with no subnet mask
       type: string
-      pattern: '^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$'
+      format: ipv4
       example: "84.125.93.10"
+
+    SingleIpv6Address:
+      description: A single IPv6 address with no subnet mask
+      type: string
+      format: ipv6
+      example: 2001:db8:85a3:8d3:1319:8a2e:370:7344
 
     Ipv4Address:
       type: string
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -986,7 +986,6 @@ components:
 
     ApplicationServerIpv4Address:
       type: string
-      pattern: '^([01]?\d\d?|2[0-4]\d|25[0-5])(?:\.(?:[01]?\d\d?|2[0-4]\d|25[0-5])){3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$'
       example: "192.168.0.1/24"
       description: |
         IPv4 address may be specified in form <address/mask> as:
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -913,7 +913,7 @@ components:
         ipv4Address:
           $ref: "#/components/schemas/DeviceIpv4Addr"
         ipv6Address:
-          $ref: "#/components/schemas/SingleIpv6Address"
+          $ref: "#/components/schemas/DeviceIpv6Address"
       minProperties: 1
 
     ApplicationServer:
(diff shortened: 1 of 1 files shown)


hdamker:
See my [comment](https://github.com/camaraproject/QualityOnDemand/issues/230#issuecomment-1818992126) within the issue.

eric-murray:
MegaLinter is complaining about trailing spaces again

eric-murray (approved):

sfnuser (approved):

hdamker (approved):
LGTM, thanks @jlurien for the update

RandyLevensalor:
@jlurien @eric-murray I thought that this was just a change to the device and not the application server.

We have use cases for supporting subnets for the application server, including 0.0.0.0/0 for all ipv4 traffic from a specific device and similarly for ipv6.

eric-murray:
> @jlurien @eric-murray I thought that this was just a change to the device and not the application server.
> 
> We have use cases for supporting subnets for the application server, including 0.0.0.0/0 for all ipv4 traffic from a specific device and similarly for ipv6.

The modified application server schema still allows for the specification of subnets, including /0 subnets

jlurien:
I forgot to remark that I have kept the pattern for ApplicationServerIpv4Address. I think that one is quite safe, but as it is the only one with pattern it may appear as inconsistent with the other definitions. I have doubts here.

hdamker:
> I forgot to remark that I have kept the pattern for ApplicationServerIpv4Address. I think that one is quite safe, but as it is the only one with pattern it may appear as inconsistent with the other definitions. I have doubts here.

My vote would go for eliminating the pattern as in `ApplicationServerIpv6Address` and stay with `string` and the description.

BTW: I will finalize the release candidate PR #236 under the assumption that we are closing this PR here. At latest in our call on Friday that should get a go.

hdamker:
> > > I forgot to remark that I have kept the pattern for ApplicationServerIpv4Address. I think that one is quite safe, but as it is the only one with pattern it may appear as inconsistent with the other definitions. I have doubts here.
> > 
> > 
> > My vote would go for eliminating the pattern as in `ApplicationServerIpv6Address` and stay with `string` and the description.
> > BTW: I will finalize the release candidate PR #236 under the assumption that we are closing this PR here. At latest in our call on Friday that should get a go.
> 
> For ApplicationServerIp**v6**Address, pattern is removed. But there is a pattern also in ApplicationServerIp**v4**Address.

My "vote" was for removing the pattern also for ApplicationServerIp**v4**Address. But it don't harm therefore I'm also fine with keeping it, at least for now.

hdamker:
@RandyLevensalor: can you comment/approve if your use case are still covered with the answers above, especially with the [one](https://github.com/camaraproject/QualityOnDemand/pull/237#issuecomment-1830704273) from Eric? If yes we would be good to go here.

jlurien:
> > > > I forgot to remark that I have kept the pattern for ApplicationServerIpv4Address. I think that one is quite safe, but as it is the only one with pattern it may appear as inconsistent with the other definitions. I have doubts here.
> > > 
> > > 
> > > My vote would go for eliminating the pattern as in `ApplicationServerIpv6Address` and stay with `string` and the description.
> > > BTW: I will finalize the release candidate PR #236 under the assumption that we are closing this PR here. At latest in our call on Friday that should get a go.
> > 
> > 
> > For ApplicationServerIp**v6**Address, pattern is removed. But there is a pattern also in ApplicationServerIp**v4**Address.
> 
> My "vote" was for removing the pattern also for ApplicationServerIp**v4**Address. But it don't harm therefore I'm also fine with keeping it, at least for now.

I think this is coherent approach, so implementations do not have to apply different logic per IP address version. I will remove it as well.

eric-murray:
@jlurien
Are you still planning to commit additional changes to remove the pattern for IPv4 addresses?

hdamker (approved):

hdamker:
Counting also Eric's previous approval in, we are done. I merge now to be able to prepare the RC for our call today.

Issue #166, opened 2023-06-16
Extend QoS Profile queries to list profiles based on specific users or devices


RandyLevensalor:
**Problem description**
Some QoS profiles may not be available to every customer.  This can be due to their current plan, capabilities/capacity of the local network, device features or other attributes.  So only a set of available profiles should be returned.

**Possible evolution**

**Alternative solution**

**Additional context**

hdamker:
MOM-2023-12-01:
- Potential candidate for v0.11.0
- Adding another endpoint with device parameter in body of POST operation could be one option

emil-cheung:
> **Problem description** Some QoS profiles may not be available to every customer. This can be due to their current plan, capabilities/capacity of the local network, device features or other attributes. So only a set of available profiles should be returned.
> 
> **Possible evolution**
> 
> **Alternative solution**
> 
> **Additional context**

The discovery endpoint filters QoS profiles based on target user's mobile subscription which makes a lot of sense.
However, filtering QoS profiles bases on network capacity is a bit tricky. As network capacity is so dynamic, usually the Policy node in a network shall consider.

I would suggest that we only scope the mobile subscription aspect at least at this stage.

jlurien:
Once we proceed with the agreed split (#265), we can discuss further enhancements.
I see two approaches:

1)  If we consume that endpoint with client credentials, we'd need a filter to be passed. If we have to pass personal information as phone numbers, we probably need to change to a POST method.
2) With a three-legged access token we can rely on the info linked to the access token, to let the implementation adapt the response to it.

RandyLevensalor:
This is looking similar to the query in issue #101 and I think that it could be follow a similar pattern using both the three-legged access token and the device schema in the request.

jlurien:
Yes, it's quite the same situation

PR #255, opened 2024-01-12
Updating the project scope in the ReadMe

### LINKED ISSUE #238: Remove mobile from QoD scope
**Problem description**
The current [scope](https://github.com/camaraproject/QualityOnDemand/blob/1ebca6f7ae9eb37bcbc627007f548c1f5a50f472/README.md?plain=1#L14) limits the API to `set quality for a mobile connection`

This contradicts the commonalities [documentation](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#25-reduce-telco-specific-terminology-in-api-definitions) to not use telecommunications specific terms. 

The documentation in the [API definition](https://github.com/camaraproject/QualityOnDemand/blob/1ebca6f7ae9eb37bcbc627007f548c1f5a50f472/code/API_definitions/qod-api.yaml#L5) also conaties the mobile specific contents.

> The Quality-On-Demand (QoD) API provides programmable interface for developers and other users (capabilities consumers) to request stable latency or throughput managed by Telco networks without the necessity to have an in-depth knowledge of the 4G/5G system or the overall complexity of the Telecom Systems.

**Expected action**

`set quality for an access network connection` or something similar that would apply to mobile and wireline networks.

Update the documentation in the API definition to be consistent with the wording.
 
**Additional context**

RandyLevensalor:
Updating the project scope to be inclusive of wire-line networks.

#### What type of PR is this?

* documentation

Updates the subproject scope to include wire-line networks in additional to mobile networs.  This reflects the state of the API as well as current operatos working on this subproject.

Fixes #238

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```


README.md
```diff
@@ -9,12 +9,13 @@
 Repository to describe, develop, document and test the QualityOnDemand API family
 
 ## Scope
-* Service APIs for “Quality on Demand” (see APIBacklog.md)  
+
+* Service APIs for “Quality on Demand” (see APIBacklog.md)
 * It provides the customer with the ability to:  
-  * set quality for a mobile connection (e.g. required latency, jitter, bit rate)  
+  * set quality for access network connections (e.g. required latency, jitter, bit rate)
```
hdamker, line 15:
```suggestion
  * set quality for access network connections (e.g. mobile device connection or fixed access towards a home gateway)
  * set the quality to QoS profiles provided by the network operator (with e.g. certain latency, jitter and bit rate parameters)  
```

some more clarity, that it isn't possible to require arbitrary parameters.

eric-murray:
@RandyLevensalor

Is this your final proposal? There is an outstanding suggested change from @hdamker. If you don't want to incorporate that, can you comment why and dismiss the suggested change.

README.md
RandyLevensalor: I'll incorporate this feedback, but would like to use something other than `towards a home gateway,` since this could be interpreted as only downstream flows.  What about `fixed access between a home gateway and the service providers gateway router`?

hdamker: @RandyLevensalor get your point and totally fine with your proposal

jlurien:
I've been checking this issue internally and it is not as simple as changing the README. It has implications on the scope of the API, as this API was originally proposed as an API for mobile networks when presented to the API Backlog. Even if most of concepts and terms introduced in the API would be applicable to fixed networks as well, implementation and commercialization of the same API covering both technologies is not that straight-forward. AFAIK this topic is also being discussed in GSMA Product stream, so I suggest to keep this on hold until some consensus is achieved across Product owners, or raise it to CAMARA Backlog as a new feature to include in the CAMARA porfolio. cc @jgarciahospital

RandyLevensalor:
@jlurien thanks for the feedback

> Even if most of concepts and terms introduced in the API would be applicable to fixed networks as well, implementation and commercialization of the same API covering both technologies is not that straight-forward. 

We have implemented this API in our lab across DOCSIS, PON and 5G networks from the same API endpoint.  Once we added the QoS Profiles, it became much easier to do this.

Once we resolve issue #166, it will also help to identify which QoS profiles are supported by a given device.  That will help filter QoS profiles based on network and device capabilities.

I don't see any technical issues with this implementation across fixed and mobile networks.  Can you expand an the none technical issues?  I'd be happy to join any discussions to share our experience.

jlurien:
Hi @RandyLevensalor, the debate is not about the technical feasibility and complexity, but about the convenience to address both mobile connections and fixed connections in the same API. CAMARA APIs are intended to be offered by different operators in a consistent way, so consensus about the scope of an API has to be reached and any subsequent change has to be discussed in the forums where Product teams are present.

RandyLevensalor:
@jlurien Thanks for the clarification.

We see enabling a single API as critical for both wireless-wireline convergence (WWC) and a better mobile Wi-Fi experience.  Additionally, the growth of fixed wireless access (FWA) is further blurring the lines between mobile and wireline networks.

Can you please share where these are being discussed in CAMARA or GSMA either on this thread or a DM in slack / e-mail if there is GSMA member only content that shouldn't be shared publicly?  I'd really like to have the opportunity to participate and/or have a representative for the wireline group on some operators provide input.

hdamker:
@jlurien @RandyLevensalor 

The functional scope of CAMARA includes both mobile and fixed networks of Telco operators ([ProjectCharter.md](https://github.com/camaraproject/Governance/blob/main/ProjectCharter.md#scope-of-the-project)):

> From a functional perspective the scope is limited to telco APIs, that means APIs in the domain of telco mobile networks, telco fixed line networks, telco edge cloud, etc. or supporting these (e.g. for authentication).

The scope of QoD was only initially limited to mobile (4G/5G). But we had already spent effort within 0.9.0 and 0.10.0 to get/keep the API technically suitable for both mobile and fixed network (see e.g. the discussion about parameters in QoSProfiles). If there are no technical reasons to restrict the scope of the API still to mobile networks we should decide to raise this limit. That was also the consensus within the last QoD call.

That the technical scope of the API includes fixed networks does not imply that any operator has to offer a product based on the API for its fixed network. The definition of products and the implementation mechanism behind the API are out of scope of CAMARA. 

BTW: Due to project history there is no proposal within API Backlog for QoD which need to be changed/updated. It was therefore also never presented within the API Backlog group.

README.md
```diff
@@ -6,23 +6,27 @@
...
-* Service APIs for “Quality on Demand” (see APIBacklog.md)  
+
+* Service APIs for “Quality on Demand” (see APIBacklog.md)
 * It provides the customer with the ability to:  
-  * set quality for a mobile connection (e.g. required latency, jitter, bit rate)  
+  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)  
   * get notification if network cannot fulfill  
-  * NOTE: The scope of this API family should be limited (at least at a first stage) to 4G and 5G.  
-* Describe, develop, document and test the APIs (with 1-2 Telcos)  
+  * NOTE: The scope of this API family should be limited the networks between the users's device (e.g. mobile phone or home gateway) and the gateway from the operator's network to other networks.
```
hdamker, line 18:
```suggestion
  * NOTE: The scope of this API family should be limited to the networks between the users's device (e.g. mobile phone or home gateway) and the gateway from the operator's network to other networks.
```

Actually this "limitation" isn't any more necessary, as the same scope description is also already in line 16.

hdamker:
a) Merged recent changes from main (release v0.10.0). 
b) Added maintainers for review.

Please review before the next QoD call on March 22nd.

maxl2287 (approved):

sfnuser (approved):

hdamker (approved):
Looks that we are ready to go.

Issue #249, opened 2023-12-20
Duration in QosProfile and SessionInfo


jlurien:
**Problem description**
In the latest definition (v0.10.0-rc), we have the option to provide a duration while creating the session and there, the maximum duration of a session is limited to 86400s (1d).
We also have discovery of QoSProfiles by mean of GET /qos-profiles, and there we return per QoSProfile a minDuration and maxDuration, that is no limited. We potentially may allow sessions for up to <Int32> days.

There is a lack of consistency between both endpints.

**Possible evolution**

It makes sense to remove the explicit limitation of duration in sessionInfo, and rely on the discovery of QoSProfiles to let the client know about the allowed limits. The same schema for Duration can be aplied in both.

hdamker:
With last QoD call we have discussed to keep the limit for duration in sessionInfo within v0.10.0 and add a note for the maxDuration parameter within QoSProfile that within v0.10.0 it shouldn't exceed of 86400 seconds/1 day.

The discussion if we should remove the explicit limitation for duration in sessionInfo will be done for the next release, as well as the alignment of the schema for duration.

hdamker:
Removing 'v0.10.0' as the part for this release is temporarily fixed with #256

hdamker:
Fixed with #296

PR #256, opened 2024-01-23
Added a note to maxDuration parameter within qosProfile about the limit of 86400 seconds

hdamker:
#### What type of PR is this?

* documentation

Add a note to maxDuration parameter that the value shouldn't exceed the current limitation within sessionInfo of 86400 seconds

This is a temporary measure for #249 but won't fix it. See comment within the issue.

#### Changelog input

```
Added a note to maxDuration parameter within qosProfile about the limit of 86400 seconds

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -687,6 +687,8 @@ components:
         maxDuration:
           description: |
             The maximum time period that this profile can be deployed.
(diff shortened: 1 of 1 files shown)


jlurien (approved):

RandyLevensalor (approved):
Thanks for adding the clarification.

Issue #254, opened 2024-01-11
Send notifications for all qosStatus changes


hdamker:
**Problem description**

The documentation within v0.10.0 doesn't currently define precisly for which qosStatus changes a developer can expect to receive a notification, see the description below (and which notifications must be implemented by an QoD API provider).

The discussion topic (see below) suggested to send notifications for all changes of the status of the QoS Session resources, including the initial entering of "available" and the ones induced by API calls by the API consumer (e.g. delete of the QoS session). This behaviour was agreed within the [QoD call of Dec 15th](https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/MeetingMinutes/MOM-2023-12-15.md).

**Expected behavior**

The documentation describes the expected behaviour. Test cases are defined for all qosStatus changes.

**Alternative solution**

**Additional context**

### Discussed in https://github.com/camaraproject/QualityOnDemand/discussions/242

<div type='discussions-op-text'>

<sup>Originally posted by **shilpa-padgaonkar** December 12, 2023</sup>
As far as I can understand, if response to create session leads to qosStatus "requested", then the developer can later receive the notification, qosStatus "available". However if the create session response leads directly to qosStatus "available", the notification with status "available" will be skipped. Similarly, before explicit delete session or delete session due to timer expiry, there is no notification sent as qosStatus "unavailable".  Is this understanding correct?
If yes, what is the thought around always sending notifications whenever qosStatus becomes available or unavailable (due to whatever conditions).
</div>

hdamker:
This issue will get fixed by #258

hdamker:
#258 is merged and part of v0.10.0-rc2

PR #258, opened 2024-01-23
Added clarification of notifications to be sent and added StatusInfo "DELETE_REQUESTED" to addres #241

### LINKED ISSUE #241: Extend the QoD API documentation with details as to what happens when qosStatus changes from available to unavailable
**Problem description**
In the documentation, https://github.com/camaraproject/QualityOnDemand/blob/main/documentation/SupportingDocuments/Presentations/QoS%20Status%20and%20Corresponding%20Notification%20Events.pdf, in the flow picture titled "Camara layer logic" it seems like - after the qosStatus changes at step 11 from "Available" to "Unavailable" (not because of duration expiry), we do not terminate the QoS session. This fact is not specified explicitly in the current API documentation clearly. Could we extend the documentation with this point so that it is also clear to the developers. This will then make the developers aware that they need to explicitly invoke a delete session. after getting a notification that qosStatus is unavailable.

**Expected action**
Extend the documentation with the above said missing content.

**Additional context**

hdamker:
#### What type of PR is this?

* documentation
* correction

Adding clarification that notification will be sent for all changes of qosStatus and what happens when qosStatus change from AVAILABLE to UNAVAILABLE

Fixes #241 and #254

As notification should be sent in all QosStatus changes, there was the need to add the StatusInfo "DELETE_REQUESTED". 

#### Changelog input

```
Clarified in documentation that notifications will be sent for all changes of QosStatus, even if initiated by the client.
Clarified in documentation that the QoS session resource what will happen when qosStatus changes from 'AVAILABLE' to 'UNAVAILABLE' due to 'NETWORK_TERMINATED'

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -105,10 +105,10 @@ paths:
         A `QOS_STATUS_CHANGED` event notification with `qosStatus` as `UNAVAILABLE` will also be send if the network terminates the session before the requested duration expired
 
         NOTE: in case of a `QOS_STATUS_CHANGED` event with `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED` the resources of the QoS session
-        are not directly set free, but will get deleted automatically only 360 seconds after the event.
+        are not directly released, but will get deleted automatically only 360 seconds after the event.
         This behavior should allow clients which are not receiving notification events but are polling the session information the opportunity
         to see the `qosStatus` `UNAVAILABLE` and `statusInfo` `NETWORK_TERMINATED`. Before a client can attempt to create a new QoD session
-        for the same device and flow within this time period they must free the session resources with an explicit `delete` operation.
+        for the same device and flow within this time period they must release the session resources with an explicit `delete` operation.
 
       operationId: createSession
       requestBody:
@@ -291,7 +291,7 @@ paths:
         - QoS Sessions
       summary: Delete a QoS session
       description: |
-        Free resources related to QoS session
+        Release resources related to QoS session
 
         If the notification callback is provided and the `qosStatus` of the session was `AVAILABLE` the client will receive in addition to the response a `QOS_STATUS_CHANGED` event with
         - `qosStatus` as `UNAVAILABLE` and
@@ -1060,8 +1060,8 @@ components:
     EventQosStatus:
       description: |
         The current status of a requested or previously available session. Applicable values in the event are:
-        *  `AVAILABLE` - The requested QoS has been provided by the network. NOTE: this event will be sent always, also if status within the response to `createSession` is already `AVAILABLE`
-        *  `UNAVAILABLE` - A requested or previously available QoS session is currently unavailable. `statusInfo` may provide additional information about the reason for the unavailability. NOTE: this event will be sent always, also if the user deletes the session explicitly
+        *  `AVAILABLE` - The requested QoS has been provided by the network.
+        *  `UNAVAILABLE` - A requested or previously available QoS session is now unavailable. `statusInfo` may provide additional information about the reason for the unavailability.
       type: string
       enum:
         - AVAILABLE
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -95,11 +95,21 @@ paths:
       description: |
         Create QoS Session to manage latency/throughput priorities
 
-        If the notification callback is provided the client will receive in addition to the response a `QOS_STATUS_CHANGED` event with either
-        - `qosStatus` as `AVAILABLE` when the session has been provided by the network, or
-        - `qosStatus` as `UNAVAILABLE` if the requested session is currently not available.
+        If the qosStatus in the API response is "AVAILABLE" and a notification callback is provided the client will receive in addition to the response a
+        `QOS_STATUS_CHANGED` event notification with `qosStatus` as `AVAILABLE`.
 
-        If the `qosStatus` in the API response is `REQUESTED`, the notification will be sent either after the network notifies that it has created the requested session or after the network notifies that it has failed to provide the requested session.
+        If the `qosStatus` in the API response is `REQUESTED`, the client will receive either
+        - a `QOS_STATUS_CHANGED` event notification with `qosStatus` as `AVAILABLE` after the network notifies that it has created the requested session, or
+        - a `QOS_STATUS_CHANGED` event notification with `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED` after the network notifies that it has failed to provide the requested session.
+
+        A `QOS_STATUS_CHANGED` event notification with `qosStatus` as `UNAVAILABLE` will also be send if the network terminates the session before the requested duration expired
+
+        NOTE: in case of a `QOS_STATUS_CHANGED` event with `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED` the resources of the QoS session 
+        are not directly set free, but will get deleted automatically only 360 seconds after the event.
+        This behavior should allow clients which are not receiving notification events but are polling the session information the opportunity 
+        to see the `qosStatus` `UNAVAILABLE` and `statusInfo` `NETWORK_TERMINATED`. Before a client can attempt to create a new QoD session 
+        for the same device and flow within this time period they must free the session resources with an explicit `delete` operation.
+
       operationId: createSession
       requestBody:
         description: Parameters to create a new session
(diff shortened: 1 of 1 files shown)


code/API_definitions/qod-api.yaml
```diff
@@ -273,7 +280,12 @@ paths:
       tags:
         - QoS Sessions
       summary: Delete a QoS session
-      description: Free resources related to QoS session
+      description: |
+        Free resources related to QoS session
+
+        If the notification callback is provided the client will receive in addition to the response a `QOS_STATUS_CHANGED` event with
+        - `qosStatus` as `UNAVAILABLE` and
+        - `statusInfo` as `USER_DELETED`
```
hdamker, line 288:
```suggestion
        - `statusInfo` as `DELETE_REQUESTED`
```

jlurien:
In my understanding, the PR is not directly addressing the issue description:

> - after the qosStatus changes at step 11 from "Available" to "Unavailable" (not because of duration expiry), we do not terminate the QoS session. This fact is not specified explicitly in the current API documentation clearly. Could we extend the documentation with this point so that it is also clear to the developers. **This will then make the developers aware that they need to explicitly invoke a delete session. after getting a notification that qosStatus is unavailable.** 

We are not indicating that the UNAVAILABLE notification does not delete the session automatically. The PR addresses that notifications will be sent even after the session is deleted by the user. But in case of unavailability due to network terminated, this is another case.

hdamker:
@jlurien You are right. I have only addressed yet the derived conclusion from our discussion in the call on Dec 15th (added to the issue):

> Design target is to allow developers to decide between notifications and polling, both methods should be sufficient to get all important information

I will add the description of the solution which we discussed for the status change due to "NETWORK_TERMINATED".

hdamker:
Added the behavior in case of `NETWORK_TERMINATED` and the need to delete a session explicitly after such event with the time period of 360 seconds before a new QoS session for the same device and flow can be created.

In addition clarified that there will be no notification event if a session with qosStatus `UNAVAILABLE` will be deleted.

hdamker:
@jlurien @shilpa-padgaonkar @maxl2287 @RandyLevensalor @eric-murray would be great if you could review this PR (again) before the QoD call tomorrow. It would be good if we can decide tomorrow about a second release candidate of v0.10.0 which includes this PR.

shilpa-padgaonkar (dismissed):
/LGTM

code/API_definitions/qod-api.yaml
```diff
@@ -92,7 +92,24 @@ paths:
       tags:
         - QoS Sessions
       summary: Creates a new session
-      description: Create QoS Session to manage latency/throughput priorities
+      description: |
+        Create QoS Session to manage latency/throughput priorities
+
+        If the qosStatus in the API response is "AVAILABLE" and a notification callback is provided the client will receive in addition to the response a
+        `QOS_STATUS_CHANGED` event notification with `qosStatus` as `AVAILABLE`.
```
maxl2287, line 99: From a developer's point of view, I'm unsure how it can be possible for the QoD-Application to directly respond with 'AVAILABLE' after creating the session. 
According to the [3GPP standard for using '3gpp-as-session-with-qos'](https://www.3gpp.org/ftp/specs/archive/OpenAPI/Rel-16/TS29122_AsSessionWithQoS.yaml), the response of a created subscription does not immediately provide information about ,e.g., 'SUCCESSFUL_RESOURCES_ALLOCATION.' 
This information is typically delivered through a callback, as far as I know.

So, in my understanding, after the session creation is successful, the qosStatus can only be 'REQUESTED.'

What do you think, @hdamker?

RandyLevensalor: I agree that in general this would not be implemented as a none blocking call, so most, if not all implementations would not use `SUCCESSFUL_RESOURCES_ALLOCATION`.

Including this as an optional return code won't break anything. Since never returning it allowed.

code/API_definitions/qod-api.yaml
```diff
@@ -273,7 +290,13 @@ paths:
       tags:
         - QoS Sessions
       summary: Delete a QoS session
-      description: Free resources related to QoS session
+      description: |
+        Free resources related to QoS session
```
RandyLevensalor, line 294: What about "Release resources" instead of "Free resources?" Since free can have several meanings, which could be ambiguous.

code/API_definitions/qod-api.yaml
```diff
@@ -1034,8 +1060,8 @@ components:
     EventQosStatus:
       description: |
         The current status of a requested or previously available session. Applicable values in the event are:
-        *  `AVAILABLE` - The requested QoS has been provided by the network
-        *  `UNAVAILABLE` - A requested or previously available QoS session is currently unavailable. `statusInfo` may provide additional information about the reason for the unavailability.
+        *  `AVAILABLE` - The requested QoS has been provided by the network. NOTE: this event will be sent always, also if status within the response to `createSession` is already `AVAILABLE`
```
RandyLevensalor, line 1063: I don't understand this note.

Does this mean that "AVAILABLE will always be sent if the status of `createSession` is also `AVAILABLE`"

code/API_definitions/qod-api.yaml
```diff
@@ -1034,8 +1060,8 @@ components:
...
+        *  `UNAVAILABLE` - A requested or previously available QoS session is currently unavailable. `statusInfo` may provide additional information about the reason for the unavailability. NOTE: this event will be sent always, also if the user deletes the session explicitly
```
RandyLevensalor, line 1064: This note is unclear.

Would this be `UNAVAILABLE` after the user has successfully deleted a session?

RandyLevensalor (changes requested):
Please see comments below regarding the notes below.

code/API_definitions/qod-api.yaml
jlurien: Here we are not creating a subscription but a QoS session with implicit notifications. We are always assuming that session is synchronously created (201) even if QoS may not be immediately available. The problem with this approach is that complicates billing for QoS sessions, as only the part of the session with QoS available should be billed.

hdamker: @maxl2287 The API should be technology agnostic. While an implementation against the 3GPP SCEF/NEF standard will - as you described - always return initially 'REQUESTED' (which the API allows), implementations of qosProfiles against other technologies might be immediately 'AVAILABLE' after the creation of the QoS session resource (e.g. if using pre-allocated resources or just prioritisation).

hdamker: > We are always assuming that session is synchronously created (201) even if QoS may not be immediately available. The problem with this approach is that complicates billing for QoS sessions, as only the part of the session with QoS available should be billed.

@jlurien My understanding it here a little bit different: there is the "QoD session resource" which will be created synchronously, and there is the actual network QoS session which get's 'available' and 'unavailable'. The ability to bill this correctly is one of the reason to create the respective events as described within this PR.

code/API_definitions/qod-api.yaml
maxl2287: The Notification-Event (CloudEvent) will include a Status Information about "UNAVAILABLE", which will be send to the consumer after deletion.

maxl2287 (approved):

maxl2287:
LGTM

code/API_definitions/qod-api.yaml
hdamker: I agree, "release" is the better choice also for me. I haven't introduced it with this PR, but instead used it from here also in two other occurrences. Replaced with "release" in all three occurrences now.

code/API_definitions/qod-api.yaml
hdamker: I agree that this note is not understandable. I've deleted it as the (to be) expected behaviour is already described within the documentation of 'createSession'.

code/API_definitions/qod-api.yaml
hdamker: I agree that this note is not understandable. I've deleted it as the (to be) expected behaviour is already described within the documentation of 'delete'.

code/API_definitions/qod-api.yaml
jlurien: @hdamker My understanding for what we have right now is the same. My comment is a reflection about whether our current approach is what consumers will expect or understand, and the implications that the current model may have on QoS session billing, after some feedback I recently received. Client will ask for a QoS session of 1h which counts from the creation (201) until expiration or deletion, but QoS is only effectively provided between status=AVAILABLE and status=UNAVAILABLE. But this is probably a discussion for another issue.

code/API_definitions/qod-api.yaml
hdamker: Done.

code/API_definitions/qod-api.yaml
hdamker: Done

code/API_definitions/qod-api.yaml
hdamker: Done

hdamker:
As agreed within our call I've added the clarification that the automatic deletion of session resources after `NETWORK_TERMINATED` will happen at earliest after 360 seconds (but could happen also later).

@jlurien @maxl2287 @RandyLevensalor @eric-murray Please do another review/approval early within the week.

RandyLevensalor (approved):
LGTM

eric-murray (approved):

PR #259, opened 2024-01-23
New example for DurationOutOfRangeForQoSProfile

### LINKED ISSUE #247: Clarification: Behaviour when requested duration < minDuration
**Problem description**

Question I get from our dev team:

Currently the minimum duration for a session in the spec is just 1 second, but GET /qos-profiles allow to return a minDuration per profile. Behaviour when a value lower than minDuration is requested is not documented.

**Possible evolution**

Some options I see:

a) Implementation will set duration to minDuration, and return it in the response, in cases when input duration is lower than that value.

b) Implementations will return a 400 error with a dedicated message.

This is kind of scenarios that need to be modeled in the test plan.

jlurien:
#### What type of PR is this?

* enhancement/feature

- Added new example for DurationOutOfRangeForQoSProfile
- Version set to 0.10.0-wip

Fixes #247

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -66,7 +66,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.10.0-rc
+  version: 0.10.0-wip
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/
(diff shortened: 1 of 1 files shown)


code/API_definitions/qod-api.yaml
```diff
@@ -206,6 +206,12 @@ paths:
                     status: 400
                     code: OUT_OF_RANGE
                     message: "Invalid port ranges specified: devicePorts"
+                DurationOutOfRangeForQoSProfile:
+                  summary: The requested duration is out of the allowed range for the specific QoS profile
```
hdamker, line 210:
```suggestion
                  summary: The requested duration is out of the allowed range for the specific QoS profile
```
To make the linter happy ...

hdamker (approved):

RandyLevensalor (approved):
Nice addition

PR #275, opened 2024-03-27
Upping the checkout action in the swagger validation to remove a warning

jimbobbennett:
#### What type of PR is this?

* cleanup

The swagger validation action was missing a name, and this is added here:

image

Becomes:

image

Also this action was using v3 of the checkout action, giving a warning suggesting an upgrade to v4.

image

This upgrades the action version to remove this warning.

No issue raised as this is a cleanup PR

None

#### Changelog input

No changeling needed - this is an internal change

### DIFF
--- a/.github/workflows/swagger-editor-validator.yml
+++ b/.github/workflows/swagger-editor-validator.yml
@@ -1,10 +1,10 @@
 ---
 name: Swagger Editor validation
 
-on:  # yamllint disable-line rule:truthy
-  pull_request:
-    branches:
-      - main
+on:  workflow_dispatch # yamllint disable-line rule:truthy
+  # pull_request:
+  #   branches:
+  #     - main
 
 jobs:
   test_swagger_editor_validator_remote:
--- a/.github/workflows/swagger-editor-validator.yml
+++ b/.github/workflows/swagger-editor-validator.yml
@@ -1,10 +1,10 @@
 ---
 name: Swagger Editor validation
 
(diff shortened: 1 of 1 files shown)


hdamker:
LGTM

But:
@RandyLevensalor @eric-murray do we need the Swagger Editor Validator still in addition to MegaLinter/Spectral? Just asking as https://github.com/camaraproject/Commonalities/tree/main/artifacts/linting_rules does not have it.

RandyLevensalor:
> LGTM
> 
> But: @RandyLevensalor @eric-murray do we need the Swagger Editor Validator still in addition to MegaLinter/Spectral? Just asking as https://github.com/camaraproject/Commonalities/tree/main/artifacts/linting_rules does not have it.

Since we include swagger editor in the documentation in the readme, I like this check to help validate that the link from our docs will work.

To that point, I think we should add https://github.com/marketplace/actions/redoc-cli-github-action for redoc as well.  If there's agreement, I can add this in a separate pr.

Maybe adding these checks and the links in the readme would be worth including in commonalities.

RandyLevensalor (approved):

hdamker:
> To that point, I think we should add https://github.com/marketplace/actions/redoc-cli-github-action for redoc as well. If there's agreement, I can add this in a separate pr.

This action is not a pure check, but creates a static HTML file, if I read it correctly.

But anyway, we should open an issue if we want to discuss that further, here we are ready to merge, I suggest.

hdamker (approved):

Issue #272, opened 2024-03-11
Maximum duration 86400 in SessionInfo cannot support 'Extend Session Duration'


emil-cheung:
**Problem description**
QoD has introduced 'Extend Session Duration' since v0.10.0 which means that the total duration of a QoD session may exceed 86400.

The responses of:
 - GET /sessions/{sessionId}
 - POST /sessions/{sessionId}/extend
 
shall return the total duration of the QoD session.

However, in the datatype 'SessionInfo' used by the responses above, the max limit of 86400 remains. 

**Expected behavior**
Remove the max limit of 86400 in datatype 'SessionInfo'.

**Alternative solution**
Or define another max limit larger than 86400 in datatype 'SessionInfo'.

**Additional context**
The documentation of 'Extend Session Duration' operation:
_"Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the new remaining duration will be the maximum limit."_

hdamker:
@emil-cheung Currently the description you have cited does not allow to extend beyond the "maximum limit" which is defined as 86400 seconds in `SessionInfo`:

> The remaining duration plus the requested additional duration shall not exceed the maximum limit. **Otherwise, the new remaining duration will be the maximum limit.**

But there is another reason to remove the "86400 seconds" limit from `SessionInfo`, see #249. The `maxDuration` will be given by the QosProfile going forward (currently we have restricted `maxDuration`with a comment in the description also to 86400 seconds, to avoid inconsistencies). `maxDuration` will be then the upper limit for a session duration and "extend session duration" shall not allow to extend beyond this limit.

Do you agree to close your issue and continue the discussion within #249?

emil-cheung:
> @emil-cheung Currently the description you have cited does not allow to extend beyond the "maximum limit" which is defined as 86400 seconds in `SessionInfo`:
> 
> > The remaining duration plus the requested additional duration shall not exceed the maximum limit. **Otherwise, the new remaining duration will be the maximum limit.**
> 
[...]

PR #277, opened 2024-04-02
Fix maximum duration in session info and improve documentation

emil-cheung:
#### What type of PR is this?

* bug
* documentation

- Improve the documentation of "Extend the duration of an active session".
- Improve the datatype "SessionInfo" to remove the maximum limit of duration.

Fixes https://github.com/camaraproject/QualityOnDemand/issues/272

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -334,14 +334,14 @@ paths:
         - QoS Sessions
       summary: "Extend the duration of an active session"
       description: |
-        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration.  
+        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration.
         The new remaining duration shall not exceed the maximum duration limit. Otherwise, the new remaining duration will be set to the maximum limit. The remaining duration is calculated as the 'expiresAt' minus the current time of the extendDuration API call.
-        An example: A QoD session was orginally created with duration 80,000 seconds. 10,000 seconds later, the deveoper wants to extend the session with 20,000 seconds. 
+        An example: A QoD session was orginally created with duration 80,000 seconds. 10,000 seconds later, the deveoper wants to extend the session with 20,000 seconds.
         - Original duration: 80,000 seconds
         - Elapsed time: 10,000 seconds
         - Remaining duration: 70,000 seconds
         - New remaining duration: 86,400 seconds
-        - New duration: 96,400 seconds 
+        - New duration: 96,400 seconds
       operationId: extendQosSessionDuration
       parameters:
         - name: sessionId
--- a/code/API_definitions/qod-api.yaml
+++ b/code/API_definitions/qod-api.yaml
@@ -334,7 +334,16 @@ paths:
         - QoS Sessions
       summary: "Extend the duration of an active session"
       description: |
-        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the new remaining duration will be the maximum limit.
+        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration.  
+        
+        The new remaining duration shall not exceed the maximum duration limit. Otherwise, the new remaining duration will be set to the maximum limit. The remaining duration is calculated as the 'expiresAt' minus the current time of the extendDuration API call.  
+
+        An example: A QoD session was orginally created with duration 80,000 seconds. 10,000 seconds later, the deveoper wants to extend the session with 20,000 seconds. 
+        - Original duration: 80,000 seconds
+        - Elapsed time: 10,000 seconds
+        - Remaining duration: 70,000 seconds
+        - New remaining duration: 86,400 seconds
+        - New duration: 96,400 seconds 
       operationId: extendQosSessionDuration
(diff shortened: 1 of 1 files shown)


RandyLevensalor (changes requested):

code/API_definitions/qod-api.yaml
```diff
@@ -334,7 +334,14 @@ paths:
         - QoS Sessions
       summary: "Extend the duration of an active session"
       description: |
-        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration. The remaining duration plus the requested additional duration shall not exceed the maximum limit. Otherwise, the new remaining duration will be the maximum limit.
+        Extend the duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additional duration.
+        The new remaining duration shall not exceed the maximum duration limit. Otherwise, the new remaining duration will be set to the maximum limit. The remaining duration is calculated as the 'expiresAt' minus the current time of the extendDuration API call.
+        An example: A QoD session was orginally created with duration 80,000 seconds. 10,000 seconds later, the deveoper wants to extend the session with 20,000 seconds.
```
jlurien, line 339:
```suggestion
        An example: A QoS session was originally created with duration 80,000 seconds. 10,000 seconds later, the developer wants to extend the session with 20,000 seconds.
```

code/API_definitions/qod-api.yaml
```diff
@@ -334,7 +334,14 @@ paths:
...
+        - Original duration: 80,000 seconds
+        - Elapsed time: 10,000 seconds
+        - Remaining duration: 70,000 seconds
+        - New remaining duration: 86,400 seconds
```
jlurien, line 343:
```suggestion
        - New remaining duration: 86,400 seconds, which is the maximum duration limit for a session
```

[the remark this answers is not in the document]
jlurien: It is right, as 86400 is the current limit maximum duration limit for a session.  
If we remove this limitation for API spec and rely on QoS profiles discovery, we would have to amend the example. We may also decide now to remove the example, even if it is clarifying,  and rely on the general description.

jlurien (commented):
Some corrections

code/API_definitions/qod-api.yaml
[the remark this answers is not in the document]
hdamker: Agree here with @jlurien that the example is correct if the limit for (remaining) duration is 86400. 

For clarification we should maybe add as a first example one which will be within the max duration range. Or omit the example completely as proposed as an option by @jlurien.

code/API_definitions/qod-api.yaml
eric-murray: The term "maximum duration limit" is ambiguous, because we have two separate limits:
- The maximum **initial** duration when the session is created
- The maximum **remaining** duration if the session is extended

Currently, both limits are set to 86,400 seconds, We may intend that they both always have the same value, but they do not have to have the same value, and the documentation should clarify that they are logically different parameters.

I'll suggest some alternative text to clarify which is intended here.

hdamker: > I'll suggest some alternative text to clarify which is intended here.

@eric-murray Which text you are suggesting?

code/API_definitions/qod-api.yaml
```diff
@@ -334,7 +334,14 @@ paths:
...
+        - New duration: 96,400 seconds
```
eric-murray, line 344:
```suggestion
        Extend the overall duration of an active QoS session. If this operation is executed successfully, the new duration of the target session will be the original duration plus the additionally requested duration.
        The new remaining duration of the QoS session shall not exceed the maximum remaining duration limit (currently fixed at 86,400 seconds) where the remaining duration is calculated as the difference between the `expiresAt` and current time when the request to extend the session duration is received. If this maximum limit would be exceeded, the overall duration shall be set such that the remaining duration is equal to this limit.
        An example: A QoD session was originally created with duration 80,000 seconds. 10,000 seconds later, the developer requested to extend the session by 20,000 seconds.
        - Original duration: 80,000 seconds
        - Elapsed time: 10,000 seconds
        - Remaining duration: 70,000 seconds
        - New remaining duration: 86,400 seconds (the maximum allowed)
        - New overall session duration: 96,400 seconds
```

code/API_definitions/qod-api.yaml
eric-murray: I suggested some alternative text below.

code/API_definitions/qod-api.yaml
hdamker: @emil-cheung the suggested text from @eric-murray covered the open points which we discussed in our QoD meeting last Friday. Assuming that the changes are also for you, I've committed them, so that we can start the final review.

code/API_definitions/qod-api.yaml
hdamker: Was covered by Eric's proposal as well.

code/API_definitions/qod-api.yaml
hdamker: @eric-murray thanks a lot, have committed your suggestion.

hdamker:
@eric-murray @RandyLevensalor @jlurien @akoshunyadi Could you please do the final review/approval until Wednesday, April 10th? Thanks!

eric-murray (approved):

RandyLevensalor (approved):
LGTM

jlurien (approved):

hdamker (approved):

PR #294, opened 2024-05-10
Update README.md

### LINKED ISSUE #268: Provision mode for QoD
**Problem description**
QoD Service provides the customer with the ability to set ceratin profile of QoS to an access network connection. Currently, the API supoports a session mode:

* the developer wants to set the required QoD profile for a certain period of time, after which the be network configuration must be set back to the default one. 
 
But there is another possible use case for QoD, which is not currently supported:

* the developer wants to set the required QoD profile indefinitely, this is, each time that the UE connects to the network, it will use the required QoD profile instead of the default one, until the association is removed.  

**Possible evolution**

Add support for a new "provision" mode, complementary to the current "session mode", with equivalent operations under a new path, to model:

* Creating a QoD provision for a device
* Removing a QoD provision for a device
* Getting the QoD provision details for a device
* Updating QoD provision details for a device 

**Alternative solution**

Where should discuss where to model these new operations. Probably the best option is in a separate API, under the same QoD API family. Most schemas can be reused. 

**Additional context**

This new model is more convenient for those B2B use cases where the devices being connected are used for a single purpose. For example reporters on the move, using backpacks to cover events and perform live video connections, that need the networking conditions provided by the QoD profile each time the backpack is used.

jgarciahospital:
#### What type of PR is this?

* correction
* documentation

Alignment with #268 including provision mode in the APIs subgroup description

Fixes #268

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Repository to describe, develop, document and test the QualityOnDemand API famil
 * It provides the customer with the ability to:  
   * set quality for a flow within an access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)
       * Session mode, for a specific duration
-      * Provision mode, indefinitely for each time that the device connects to the same access network
+      * Provision mode, indefinitely for each time the device connects to the same access network
   * get notification if network cannot fulfill  
 * Describe, develop, document and test the APIs (with 1-2 Service Providers)  
 * Started: October 2021
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ Repository to describe, develop, document and test the QualityOnDemand API famil
 
 * Service APIs for “Quality on Demand” (see APIBacklog.md)
 * It provides the customer with the ability to:  
-  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)
+  * set quality for a flow within an access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)
       * Session mode, for a specific duration
-      * Provision mode, indefinitely for each time that the device connects to the network
+      * Provision mode, indefinitely for each time that the device connects to the same access network
   * get notification if network cannot fulfill  
 * Describe, develop, document and test the APIs (with 1-2 Service Providers)  
 * Started: October 2021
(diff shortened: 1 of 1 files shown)


eric-murray (dismissed):

README.md
```diff
@@ -13,7 +13,9 @@ Repository to describe, develop, document and test the QualityOnDemand API famil
 
 * Service APIs for “Quality on Demand” (see APIBacklog.md)
 * It provides the customer with the ability to:  
-  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)  
+  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)
+      * Session mode, for a specific duration
+      * Provision mode, indefinitely for each time that the device connects to the network
```
RandyLevensalor, line 18:
```suggestion
      * Provision mode, indefinitely for each time the device connects to the network
```

Question: Should we specific that this is just when it connects to the current network or any network.  Saying "the network" is a little ambiguous, would this apply when roaming or not?

README.md
```diff
@@ -13,7 +13,9 @@ Repository to describe, develop, document and test the QualityOnDemand API famil
 
 * Service APIs for “Quality on Demand” (see APIBacklog.md)
 * It provides the customer with the ability to:  
-  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)  
+  * set quality for access network connections (e.g. mobile device connection or fixed access between a home gateway and the service providers gateway router)
+      * Session mode, for a specific duration
```
RandyLevensalor, line 17:
```suggestion
      * Session mode, for a specific duration for the device
```

Adding "the device" to be consistent with the second bullet.

RandyLevensalor (changes requested):

README.md
hdamker: Actually is the quality set for the specified connection, not for the complete device, therefore this addition could be misleading.
My understanding is that in the second bullet "device connects to the network" is the trigger to set the quality for the (specific) connection again, but it is not set for the complete device.

RandyLevensalor (approved):

hdamker (approved):

PR #297, opened 2024-05-29
Add error 429 QUOTA_EXCEEDED

### LINKED ISSUE #257: Error behaviour when session cannot be created due to "time cap" limitation
**Problem description**

Use case would be that QoD is productized as number of N QoD minutes or sessions per period and after reaching that limit a new QoD session, or a session duration extension, is requested. What would be the right error to be answered? As an enhancement this can be added as example to document the expected behaviour.

**Possible evolution**
Add an explicit example to /paths/sessions/POST/responses/"429"/content/"application/json"/examples:

In the guidelines there is a Service Exception documented as: 

| Error Code | Description |
|---|---|
| `TOO_MANY_REQUESTS` | Either out of resource quota or reaching rate limiting. |

```
SessionsTimeCapExceeded:
  summary: The requested duration exceeds the overall sessions time cap for this device in certain period of time or contract duration.
  value:
    status: 429
    code: TOO_MANY_REQUESTS
    message: The requested session exceeds the allocated quota for this device: Sessions time cap limit exceeded
```

**Alternative solution**
An alternative code may be created especifically for `QUOTA_REACHED`.

**Additional context**

Initially opened as discussion: https://github.com/camaraproject/QualityOnDemand/discussions/231

jlurien:
#### What type of PR is this?

* enhancement/feature

Adds a new error scenario to be used when a server logic rejects a request to create a session due to a business quota being exceeded.

Added a subcase for 429 with a new specific code QUOTA_EXCEEDED

Fixes #257 

403 could have been used instead, as a general forbidden due to business rules, but 429 is more related to cases where rate or quota limits are exceeded. But this could be open to debate as guidelines are not explicit about this case. It is related with the discussion for Commonalities PR https://github.com/camaraproject/Commonalities/pull/213

#### Changelog input

```
Added explicit error response status 429 to createSessions

```

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -252,6 +252,23 @@ paths:
                 status: 409
                 code: CONFLICT
                 message: "Another session is created for the same device"
+        "429":
+          description: Rate limit or quota exceeded
+          headers:
+            x-correlator:
+              $ref: '#/components/headers/x-correlator'
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+              examples:
+                QuotaExceeded:
+                  summary: Quota exceeded
+                  description: The requested duration exceeds the overall sessions time cap for this device or application, in certain period of time or contract duration.
+                  value:
+                    status: 429
+                    code: QUOTA_EXCEEDED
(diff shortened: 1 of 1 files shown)


maxl2287:
@jlurien I have a parallel PR (without seeing here yours before) #306 
This includes also HTTP-429.
Maybe you can take a look

PR #299, opened 2024-06-05
New API QOD Provisioning

### LINKED ISSUE #268: Provision mode for QoD

jlurien:
#### What type of PR is this?

* enhancement/feature

First draft of new QoD Provision API. It is a new API in the family of Quality On Demand. It implements the new mode approved to enhance the current scope:

* Provision mode, indefinitely for each time the device connects to the same access network

Fixes #268, #332 

- Name proposal was originally "qod-provision" but it was agreed to use "qod-provisioning". Versions wip/vwip, to be changed to 0.1.0 (and its variants) for first RC

- The PR just adds the yaml spec. Other files like README, documentation, etc may need to be updated once the API is reviewed and approved.

- It follows closely the model for the sessions mode (renamed to `quality-on-demand`, final name TBD). Main differences are:
  - It does not consider `duration` and `expiresAt`, as provision is assumed to exist indefinitely, until it is revoked
  - It does not consider `applicationServer`, provisioning affects all device traffic
  - It assumes to reuses same qosProfiles than the session modes. We may review if it makes sense to have distinctions between the profiles applicable to each mode.
  - `status` replaces `qosStatus` in sessions, as it is the status of the configuration being applied, not the status of the QoS being in place at the moment. Configuration is active even if the device is temporarily not having a QoS benefit, for example because is disconnected or out of coverage. 
  - A new endpoint to retrieve the provisions for a given device is modeled. This is equivalent to the new endpoint for session in PR #325

- It already applies the latest alignment with Commonalities/ICM guidelines, and it is also aligned with #326

#### Changelog input

```
Initial version of QoD Provision mode API

```


jlurien:
From the [minutes of the last meeting](https://wiki.camaraproject.org/display/CAM/2024-06-14+Quality+on+Demand+-+Meeting+Minutes): 

> @eric-murray: Does the API still allow to have multiple provisions per device? Or should it be restricted to one, and enhance the complete traffic of the device (would simplify the API a lot)
> @jlurien : all traffic is only one possible use case (hence 'applicationServer' is proposed to be optional)
> Open: Are there use cases to provision QoD for specific flows (defined by applicationServer)? - Jose will ask on product side

I took this AP to our Product team, and Eric's suggestion makes sense for this first version.  So far, the initial use case is restricted to one provision per device, and affecting all of its traffic. Indeed, in case that it was required to include server side information for provision, we would likely need to use different identifiers for applicationServer, as server side IPs may change or they may be reused or shared by several applications (in a cloud server for example). Using AS domain names would make more sense. Even the original "session" mode may need to review these aspects in the future.

So I will simplify the original proposal accordingly. fyi @hdamker

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,876 @@
...
+          $ref: "#/components/responses/Generic404"
+        "500":
+          $ref: "#/components/responses/Generic500"
+        "503":
+          $ref: "#/components/responses/Generic503"
+      security:
+        - openId:
+            - "qod-provision:provisions:delete"
+
+  /provisions/retrieve:
```
RandyLevensalor, line 239: Would something like `/provisions/active` be a more intuitive to get the active provision on the current device.

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,876 @@
...
+                PROVISION_AVAILABLE:
+                  $ref: "#/components/examples/PROVISION_AVAILABLE"
+        "400":
+          $ref: "#/components/responses/Generic400"
+        "401":
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/RetrieveProvision404"
```
RandyLevensalor, line 276: What about a second response to 404 for device not found?  Since this code implies that there isn't an active provision, which is technically true.  There are no active provisions on a none existent device, but the implications are different.

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,876 @@
...
+  - url: "{apiRoot}/qod-provision/vwip"
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root, defined by the service provider, e.g. `api.example.com` or `api.example.com/somepath`
+tags:
+  - name: QoD Provision
+    description: Manage the provision of QoD
+paths:
+  /provisions:
```
RandyLevensalor, line 55: Is provisions the best name for this endpoint?  As I'm reviewing this, I keep thinking about getting supplies (provisions) for a camping trip.

What about `/persistent` or `/device-qod`?  Something that denotes it's setting a QoS Provide for the device and / or that it's a type of QoD session that will never expire.

RandyLevensalor (changes requested):
The more I read `/provisions`, it doesn't seem to capture the intent of this end point.  Which if I'm understanding this endpoint correctly to intent is is to apply a persistent QoS profile to a device until it is removed.

code/API_definitions/qod-provision.yaml
jlurien: I trust a native speaker on this :) We are open to change the name to whatever makes the use case more understandable and remark the difference with respect to the other API (sessions). The concept is related to the [provisioning](https://en.wikipedia.org/wiki/Provisioning_(technology)) of the QoD persistently.

For API name, maybe `qod-provisioning` is more clear? For the resource name it could be `device-qos`, and it would imply operations such as  `POST /qod-provisioning/v0.1/device-qos` and `GET /qod-provisioning/v0.1/device-qos/{device_qos_id}`

RandyLevensalor: @jlurien /qod-provisioning makes sense to me.

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,876 @@
...
+              schema:
+                $ref: "#/components/schemas/ProvisionInfo"
+        "400":
+          $ref: "#/components/responses/Generic400"
+        "401":
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/DeviceNotFound404"
```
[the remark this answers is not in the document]
jlurien: Device Not Found is the only code considered in Commonalities for 404 related to devices. For cases when the device identifiers identify more than a single device, the response 422 DEVICE_IDENTIFIERS_MISMATCH applies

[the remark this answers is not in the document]
hdamker: @jlurien what is then the right code and message if the (single) provided identifier, e.g. Phone Number, belongs to multiple devices (e.g. "MultiCard"/"MultiSIM" features)? "DEVICE_IDENTIFIERS_MISMATCH" would be misleading here ...

code/API_definitions/qod-provision.yaml
jlurien: I see that the description is wrong as it returns all provisions regardless the status. I'll have to fix that. And as it is a POST modeled to get information, the path is usually modeled as an action. A POST /resource/active would be seen as an endpoint to create a resource in active status.

code/API_definitions/qod-provision.yaml
jlurien: The response (schema in components) already considers 2 codes: PROVISION_NOT_FOUND and DEVICE_NOT_FOUND. As I said above there is an error in the description and provision could be in any status.

code/API_definitions/qod-provision.yaml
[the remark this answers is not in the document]
jlurien: @hdamker I think that this case has not been considered yet in Commonalities. It is really a problem we have from the beginning when we identify a device by a MSISDN, which may not be exclusive for a single device, as it identifies a subscriber. IMO it would be a subcase for 422, as the entity is not processable for the service. The problem is also extensible to other APIs such as the sessions mode QoD or Device Location, when a single device is expected.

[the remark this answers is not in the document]
jlurien: We still the generic 422 UNPROCESSABLE_ENTITY for this sub-case, or even DEVICE_NOT_SUPPORTED, but probably it is worthy to raise an issue to handle properly multi-sim scenarios

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,876 @@
+openapi: 3.0.3
+info:
+  title: QoD Provisioning API
+  description: |
+    The Quality-On-Demand (QoD) Provisioning API offers a programmable interface for developers to request the assignment of a certain QoS Profile to a certain device, indefinitely.
+
+    This API sets up the configuration in the network so the requested QoS profile is applied to an specified device, at any time while the provisioning is available. The device traffic will be treated with a certain QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
```
RandyLevensalor, line 7:
```suggestion
    This API sets up the configuration in the network so the requested QoS profile is applied to an specified device, at any time while the provisioning is available. The device traffic will be treated with a certain QoS profile by the network whenever the device is connected to the network, until the provisioning is deleted.

```

The description says revoked, but the API call is delete to `Release resources related to QoS provisioning.` The wording in the description should be consistent with the API.  If we need to support a case where an operator revokes the QoS provisioning on a device, then we should to be clear about that.

RandyLevensalor (changes requested):
@jlurien I'm still confused about one aspect of this end point / use case.

Since we are not specifying an application server, target port or any DCSP markings (which is a separate issue on the backlog), would this mean that all traffic would fall under this QoS profile or would the service provider just use this to update a default or predetermined flow/PDU session etc.?

If possible, could we add documentation as to how this impacts the traffic on the given device or just clarify that it's on per provider basis and may change based on type of service.  I can't imagine that we'd want to change some of the standard QoS to just flatten the profiles on the device.

hdamker:
> Since we are not specifying an application server

The PR description still mentions the optional `ApplicationServer` parameter but than was the API simplified after the discussion documented in https://github.com/camaraproject/QualityOnDemand/pull/299#issuecomment-2175809214.

I can't give a statement if this simplified API (but also the QoD Provisioning API as such) makes sense as product or not ... currently lacking feedback from product side.

code/API_definitions/qod-provision.yaml
```diff
@@ -0,0 +1,877 @@
...
+          $ref: "#/components/responses/Generic404"
+        "500":
+          $ref: "#/components/responses/Generic500"
+        "503":
+          $ref: "#/components/responses/Generic503"
+      security:
+        - openId:
+            - "qod-provisioning:device-qos:delete"
+
+  /device-qos/retrieve:
```
eric-murray, line 240: The design guidelines are a bit vague on the correct naming convention here:

> "When the POST method is used [instead of GET], the resource in the path must be a verb (e.g. `retrieve-location` and not `location`) to differentiate from an actual resource creation."

So I would expect an endpoint which would ideally be `GET /device-qos` to be named `POST /retrieve-device-qos` according to the above statement. But maybe `POST /device-qos/retrieve` also complies as `retrieve` is a verb, and I can't imagine that a developer would interpret this as creating a `retrieve`.

I'd suggest the naming convention should be clarified with Commonalities, in case both are considered to be valid options, and they can revise the guidelines accordingly.

code/API_definitions/qod-provision.yaml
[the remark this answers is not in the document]
eric-murray: @hdamker, @jlurien

If, by "multiSIM" you mean one MSISDN shared between multiple devices/SIMs (e.g. smartphone and smartwatch), this was discussed in Device Identifier, and the conclusion was that the `phoneNumber` should always be interpreted as the secondary MSISDN rather than primary.

The secondary MSISDN should be unique to a single device (SIM) - at least for the Vodafone "OneNumber" service it is, but maybe there are services where this is not true. For most devices, the secondary MSISDN will anyway be the same as the primary, and so this interpretation does not change things for such devices. The issue is that, if the secondary MSISDN is different, it is not so easy to find out what it is.

Of course, for Device Identifier this mattered, as the API is returning the IMEI of the UE, so we need to know which UE is intended. For QoS provisioning, maybe we just prioritise all devices that share the primary MSISDN.

jlurien:
> @jlurien I'm still confused about one aspect of this end point / use case.
> 
> Since we are not specifying an application server, target port or any DCSP markings (which is a separate issue on the backlog), would this mean that all traffic would fall under this QoS profile or would the service provider just use this to update a default or predetermined flow/PDU session etc.?
> 
> If possible, could we add documentation as to how this impacts the traffic on the given device or just clarify that it's on per provider basis and may change based on type of service. I can't imagine that we'd want to change some of the standard QoS to just flatten the profiles on the device.

You're right, this means that the QoD profile will be applied to all the traffic sent to or received from the line under consideration. A relevant example would be the line for a point-of-sale terminal that will be configured with a better than best-effort service so that the credit card transactions will succeed even in crowded areas.

jlurien:
The PR has been adapted to the latest decisions applied for sessions (#326, #325). PR description has been updated.

Next steps:
- Approve this PR before r1.1
- Add it as part of r1.1:
  - Add references to the new API to README, CHANGELOG 
  - Change "wip" (references in API versions, server url, and event versions) to 0.1.0-rc.1

@hdamker please let me know if you miss some other task

hdamker:
> @hdamker please let me know if you miss some other task

@jlurien merge #330 (details can be changed within the release PR, but would be good if the file is already there)

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,959 @@
...
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          examples:
+            GENERIC_404_NOT_FOUND:
+              summary: Generic Not Found
+              description: Resource is not found
+              value:
+                status: 404
+                code: NOT_FOUND
+                message: "{{resource}} is not found"
```
hdamker, line 775: The example is deviating slightly from https://github.com/camaraproject/Commonalities/blob/r0.4.0-rc.1/artifacts/CAMARA_common.yaml ... e.g. there is no `summary` and no `resources` variable.
(Same comment applies for some of the other error response examples).

```suggestion
            GENERIC_404_NOT_FOUND:
              description: Resource is not found
              value:
                status: 404
                code: NOT_FOUND
                message: The specified resource is not found.
```

jlurien: I added "summary" because it helps visualization of the examples in the viewers (e.g. Redocly), but I don't mind removing them to keep it similar everywhere. We could enhance the common artifact in a future release, in case.

Corrected also the message

jlurien:
Hi, @RandyLevensalor please review again, as you requested changes in a past review and now your explicit approval is needed to unblock the merge.

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,1061 @@
...
+
+    * **Identifier for the device**:
+    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the network operator for the device, at the request time. After the provisioning request is accepted, the device may get different IP addresses, but the provisioning will still apply to the device that was identified during the request process. Note: Network Access Identifier is defined for future use and will not be supported with v0.1 of the API.
+
+    * **Notification URL and token**:
+    Developers may provide a callback URL (`sink`) on which notifications about all status change events (eg. provisioning termination) can be received from the service provider. This is an optional parameter. The notification will be sent as a CloudEvent compliant message. If `sink` is included, it is RECOMMENDED for the client to provide as well the `sinkCredential` property to protect the notification enpoint. In the current version,`sinkCredential.credentialType` MUST be set to `ACCESSTOKEN` if provided.
```
maxl2287, line 19: typo in `notification enpoint`
Missing "d"

RandyLevensalor (approved):

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,1061 @@
...
+        End-user equipment able to connect to the network. Examples of devices include smartphones or IoT sensors/actuators.
+
+        The developer can choose to provide the below specified device identifiers:
+
+        * `ipv4Address`
+        * `ipv6Address`
+        * `phoneNumber`
+        * `networkAccessIdentifier`
+
+        NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device
```
hdamker, line 650: Same change as in https://github.com/camaraproject/QualityOnDemand/pull/326/commits/70ea70ee9386eebd4b7e102887021739cedb1385 tbd (has been changed also in Commonalities)

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,1061 @@
...
+          description: Provisioning created
+          headers:
+            x-correlator:
+              $ref: "#/components/headers/x-correlator"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProvisioningInfo"
+        "400":
+          $ref: "#/components/responses/Generic400"
```
hdamker, line 168: @jlurien Another one:

Within the event-subscription template there are two specific error codes INVALID_CREDENTIAL and INVALID_TOKEN defined for bad values in `sinkCredential`. See also https://github.com/camaraproject/QualityOnDemand/pull/335#issuecomment-2269441603

jlurien: I have added both in a new CreateProvisioning400 response, even if INVALID_TOKEN would be strictly not necessary right now, as "berarer" is the only value for `accessTokenType` allowed in the schema (enum), and sending another value would be a generic 400.

code/API_definitions/qod-provisioning.yaml
jlurien: Updated

code/API_definitions/qod-provisioning.yaml
jlurien: Thanks

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,1094 @@
...
+                      PROVISIONING_STATUS_CHANGED_EXAMPLE:
+                        $ref: "#/components/examples/PROVISIONING_STATUS_CHANGED_EXAMPLE"
+              responses:
+                "204":
+                  description: Successful notification
+                  headers:
+                    x-correlator:
+                      $ref: "#/components/headers/x-correlator"
+                "400":
+                  $ref: "#/components/responses/CreateProvisioning400"
```
hdamker, line 141:
```suggestion
                  $ref: "#/components/responses/Generic400"

```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -0,0 +1,1094 @@
...
+          description: Provisioning created
+          headers:
+            x-correlator:
+              $ref: "#/components/headers/x-correlator"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProvisioningInfo"
+        "400":
+          $ref: "#/components/responses/Generic400"
```
hdamker, line 168:
```suggestion
          $ref: "#/components/responses/CreateProvisioning400"

```

hdamker (changes requested):

code/API_definitions/qod-provisioning.yaml
hdamker: INVALID_CREDENTIAL and INVALID_TOKEN are potential responses of the operation createProvisioning, not of the callback. The callback would response with 403 if the credential is not valid.

hdamker (approved):
LGTM

hdamker:
One point which I have seen in my final review: the reference to PROVISIONING_AVAILABLE_WITHOUT_DEVICE made me curios what is behind. Learned then that it is mainly "...WITHOUT_DEVICE_INFORMATION_IN_RESPONSE". Makes sense to have this example for OAS viewers/editors.

But it brought me to a point - feel free to create an issue or discussion out of it: is the provisioning really bound to a specific device? Or - in case of phoneNumber as the device identifier - is the provisioning actually bound to a subscription? 

If the provisioning is bound to a device:
* what is the expected behaviour if currently no device with the (valid) phoneNumber is registered with the network? Will the provisioning resource getting AVAILABLE and the improved QoS established when the/a device registers with the network?
* in case the device was registered with the network when creating the provisioning ... what happens in case of a DeviceSwap with the provisioning? Should that be detected and the provisioning gets terminated by the network?

jlurien:
> One point which I have seen in my final review: the reference to PROVISIONING_AVAILABLE_WITHOUT_DEVICE made me curios what is behind. Learned then that it is mainly "...WITHOUT_DEVICE_INFORMATION_IN_RESPONSE". Makes sense to have this example for OAS viewers/editors.
> 

I just added the example to illustrate that device can be empty in responses if not provided in the request (because it was implied from the 3-legged access token), but the other questions are relevant

> But it brought me to a point - feel free to create an issue or discussion out of it: is the provisioning really bound to a specific device? Or - in case of phoneNumber as the device identifier - is the provisioning actually bound to a subscription?
> 
> If the provisioning is bound to a device:
> 
> * what is the expected behaviour if currently no device with the (valid) phoneNumber is registered with the network? Will the provisioning resource getting AVAILABLE and the improved QoS established when the/a device registers with the network?

> * in case the device was registered with the network when creating the provisioning ... what happens in case of a device swap with the provisioning? Should that be detected and the provisioning gets terminated by the network?

I will share the questions with our Product team to fully discuss the corner cases, but the "happy case" designed so far, to my knowledge, is that device identifiers (explicitly provided or implied from token) are used to identify a valid subscription (typically associated to a SIM), and provisioning is applied to it. If may happen that SIM is not connected to the network at the time of the request if a phone number is provided, and my assumption is that provisioning should be allowed, but it may depend on the implementation. Allowing IP addresses enables the possibility to identify the device in cases where phone numbers are not known by the client, as with the session mode, but if an IP address is provided the device has to be connected.

If the SIM is swapped to a different terminal it shouldn't terminate the provisioning (IMO) but again it may be business decision to terminate it, if it is detected.

There may be some implications with multi-sim scenarios, but that kind of implications may affect as well the session mode, and it depends also on how multi-sim is implemented internally by operators.

For future releases we can open a dedicated issue to discuss all of these implications.

hdamker:
@RandyLevensalor @eric-murray any last words on the PR? To complete the release PR I would need the API in `main`. For an initial version with 0.1.0 it is in quite good shape.

hdamker:
@jlurien Eric had reviewed previously and the comment is addressed, @RandyLevensalor has approved before the last change which was only in the error examples. I think we are good here - would you merge the PR, so that I can finish the Release PR?

PR #301, opened 2024-06-13
Make + prefix mandatory for phoneNumber

### LINKED ISSUE #298: Align phoneNumber with guidelines, with prefix + mandatory
**Problem description**
Current QoD versions are misaligned with latest guidelines, where E.164 is in international format starting with +:

https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml#L76

and QoD v0.10 still considers it optional:

 
**Expected behavior**

Modify pattern, description and example

jlurien:
#### What type of PR is this?

* correction

Alignment with Commonalities about + prefix for international E.164 phoneNumbers 

Fixes #298

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -775,7 +775,7 @@ components:
       type: string
       pattern: '^\+[1-9][0-9]{4,14}$'
       example: "+123456789"
-      
+
     DeviceIpv4Addr:
       type: object
       description: |
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -771,11 +771,11 @@ components:
       example: "123456789@domain.com"
 
     PhoneNumber:
-      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, optionally prefixed with '+'.
+      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
       type: string
-      pattern: '^\+?[0-9]{5,15}$'
-      example: "123456789"
-
+      pattern: '^\+[1-9][0-9]{4,14}$'
+      example: "+123456789"
+      
     DeviceIpv4Addr:
       type: object
       description: |


code/API_definitions/quality-on-demand.yaml
```diff
@@ -771,11 +771,11 @@ components:
...
     PhoneNumber:
-      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, optionally prefixed with '+'.
+      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
       type: string
-      pattern: '^\+?[0-9]{5,15}$'
-      example: "123456789"
-
+      pattern: '^\+[1-9][0-9]{4,14}$'
+      example: "+123456789"
+
```
hdamker, line 778: Make linting happy :-)

```suggestion

```

hdamker:
@jlurien looks good, but you need to make the linter happy :-)

eric-murray (approved):

code/API_definitions/quality-on-demand.yaml
```diff
@@ -771,10 +771,10 @@ components:
       example: "123456789@domain.com"
 
     PhoneNumber:
-      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, optionally prefixed with '+'.
+      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
       type: string
-      pattern: '^\+?[0-9]{5,15}$'
-      example: "123456789"
+      pattern: '^\+[1-9][0-9]{4,14}$'
```
RandyLevensalor, line 776: Why is the number of digits following the +[1-9] reduced from 5-15 to 4-14?

This change is independent of requiring the +

RandyLevensalor: I looked at commonalties and this was also requiring too many digits for the minimum and allowing too many for the max.  So fixing two issues with the phone number regex.

RandyLevensalor (approved):
This actually corrects three issues with the regex.

1. Requires +
2. Reduces the minimum length
3. Decreases max length

hdamker:
@RandyLevensalor 
> 2. Reduces the minimum length
> 3. Decreases max length

Actually there is no change of these lengths, previously it was max 15 digits of [0-9] and now it is one digit [1-9] followed by max 14 digits [0-9]. The additional check is that the first digit after the + can't be 0 (as there no country code starting with 0).

eric-murray:
@RandyLevensalor 
E.164 allows international numbers to be up to 15 digits long, so we need to support that even if there are no actual examples of 15 digit telephone numbers

jlurien:
Yes, new pattern is copied from the artifact in Commonalities

hdamker (approved):

sfnuser (approved):
Looks good to me. Thanks

Issue #313, opened 2024-07-01
Align with Commonality decision about optional device object and respective documentation


hdamker:
**Problem description**

https://github.com/camaraproject/Commonalities/pull/233 has been merged and will be part of Commonalities v0.4.0, therefore we need to adapt the quality-on-demand.yaml accordingly:

**Possible evolution**

- make device object optional
- add the recommended documentation

**Alternative solution**

Alternative documentation text if there is a need for QoD API.

**Additional context**

jlurien:
A situation that has been raised in other APIs related to allowing an optional device in the input is that our current assumption is that device is always returned as part of responses, within SessionInfo in QoD. This has impact on the implementation as server must decide which identifiers to fill in the response when no device is explicitly indicated. 

Even with our current approach of requiring an input device, it is not explicitly indicated if the implementation must reuse the same identifiers that were passed during the session creation or it may return different ones. A problematic case is when the device were identified by some IP address that may have changed. Should the implementation return the identifiers that were used during creation or should it return some other updated or permanent identifier?

hdamker:
> A situation that has been raised in other APIs related to allowing an optional device in the input is that our current assumption is that device is always returned as part of responses, within SessionInfo in QoD. This has impact on the implementation as server must decide which identifiers to fill in the response when no device is explicitly indicated.
> 
> Even with our current approach of requiring an input device, it is not explicitly indicated if the implementation must reuse the same identifiers that were passed during the session creation or it may return different ones. A problematic case is when the device were identified by some IP address that may have changed. Should the implementation return the identifiers that were used during creation or should it return some other updated or permanent identifier?

@jlurien Good point ... what is purpose or value of returning the device identifier in the session info (beyond development and testing sessions)?
* If we are returning here another identifier then provided within the request, we are risking to reveal additional personal information, that not a valid option for me
* In case that the device was identified based on the access token, there is also a risk to reveal personal information the actual API client does not need to know
 
Anyways: we should open a new issue for this, with expectation to document the result of the discussion and potential needed changes. Do you have links within the other APIs?

jlurien:
> > A situation that has been raised in other APIs related to allowing an optional device in the input is that our current assumption is that device is always returned as part of responses, within SessionInfo in QoD. This has impact on the implementation as server must decide which identifiers to fill in the response when no device is explicitly indicated.
[...]

PR #318, opened 2024-07-11
Add query to qos-profiles to query profiles available on a given device

### LINKED ISSUE #166: Extend QoS Profile queries to list profiles based on specific users or devices

RandyLevensalor:
#### What type of PR is this?

* enhancement/feature

Adds a post call to query for qos profiles based on a device.  It uses the same device paramaters as the QoS Session create call.

The original get call was not modified.

Fixes #166

Since this code was largely copied from Quality-On-Demand end point, please check for copy past errors.

#### Changelog input

```
 release-note

Added /qos-profiles-by-device post method to enable serching for QoS profiles which are available on a given device.

```

#### Additional documentation 

This section can be blank.

```
docs

```


eric-murray:
If you made `device` optional, then the existing `GET /qos-profiles` would become redundant, and a single endpoint would satisfy all use cases:
- A 3-legged token (with or without `device` specified in the body) would return all profiles available to that device
- A 2-legged token with `device` specified in the body would return all profiles available to that device
- A 2-legged token with no `device` specified in the body would return all possible profiles

RandyLevensalor:
> If you made `device` optional, then the existing `GET /qos-profiles` would become redundant, and a single endpoint would satisfy all use cases:
> 
> * A 3-legged token (with or without `device` specified in the body) would return all profiles available to that device
> * A 2-legged token with `device` specified in the body would return all profiles available to that device
> * A 2-legged token with no `device` specified in the body would return all possible profiles

@eric-murray Good points. I went back and forth on the convince of keeping the get vs forcing everything to a post.

If no one objects to just using the post, I'll make this change.

code/API_definitions/qos-profiles.yaml
```diff
@@ -47,33 +47,24 @@ tags:
...
-
-        The access token may be either a 2-legged or 3-legged access token. If the access token is 3-legged, all returned QoS Profiles must be available to all end users associated with the access token.
-
-      operationId: getQosProfiles
-      security:
-        - openId:
-            - qos-profiles:qos-profiles:read
+        Returns all QoS Profiles that match the given criteria.
+        If the client credentials specify a device, the QoS Profiles returned are those that are applicable to the device.
+        If no criteria is given and the credentials don't specify a device, all QoS Profiles are returned.
```
hdamker, line 57: "If the client credentials don't specify a device" is for me not clear enough, especially compared to the language we have in quality-on-demand and also previously here. If a 3-legged token is used the API client must only get profiles which are actually available to "all end users associated with the access token".

```suggestion
        Returns all QoS Profiles that match the given criteria.
        The access token may be either a 2-legged or 3-legged access token. If the access token is 3-legged, all returned QoS Profiles will be available to all end users associated with the access token. It
```

There is the need to add more documentation within the API description, cf #313

hdamker:
@camaraproject/quality-on-demand_maintainers within final review, please provide your comments - if any - until tomorrow, Tuesday, eob.

code/API_definitions/qos-profiles.yaml
```diff
@@ -47,33 +47,23 @@ tags:
...
-        Returns all QoS Profiles that match the given criteria, or all profiles if no criteria is specified.
-
-        The access token may be either a 2-legged or 3-legged access token. If the access token is 3-legged, all returned QoS Profiles must be available to all end users associated with the access token.
-
-      operationId: getQosProfiles
-      security:
-        - openId:
-            - qos-profiles:qos-profiles:read
+        Returns all QoS Profiles that match the given criteria.
+        The access token may be either a 2-legged or 3-legged access token. If the access token is 3-legged, all returned QoS Profiles will be available to all end users associated with the access token.
```
hdamker, line 56: Alignment of formatting with quality-on-demand

```suggestion
        **NOTES:**
        - The access token may be either a 2-legged or 3-legged access token. 
        - If the access token is 3-legged, all returned QoS Profiles will be available to all end users associated with the access token.
```

code/API_definitions/qos-profiles.yaml
```diff
@@ -377,6 +395,101 @@ components:
...
+        End-user equipment able to connect to a mobile network. Examples of devices include smartphones or IoT sensors/actuators.
+
+        The developer can choose to provide the below specified device identifiers:
+
+        * `ipv4Address`
+        * `ipv6Address`
+        * `phoneNumber`
+        * `networkAccessIdentifier`
+
+        NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device
```
hdamker, line 421: Latest version from Commonalities:

```suggestion
        NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device
        NOTE2: for the Commonalities release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the CAMARA meta-release work is concluded and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.
```

code/API_definitions/qos-profiles.yaml
```diff
@@ -86,6 +76,34 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/QosProfile"
+        "400":
```
hdamker, line 79: Please copy the missing examples from quality-on-demand.yaml in #326

hdamker (changes requested):

hdamker:
@RandyLevensalor added some of the changes which I applied in #326 to quality-on-demand here as comments.

hdamker:
@RandyLevensalor seems that there weren't further comments. Will you address my comments?

RandyLevensalor:
> @RandyLevensalor seems that there weren't further comments. Will you address my comments?

@hdamker Yes.  I'm not sure how I missed MNO being added to the comments.  So, I'll replace that mobile specific terminology with service provider to be more inclusive of wireline operators.

hdamker:
> I'm not sure how I missed MNO being added to the comments. So, I'll replace that mobile specific terminology with service provider to be more inclusive of wireline operators.

I missed that as well. And worse: it is defined with the MNO abbreviation in [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml). I will open an issue on that.

code/API_definitions/qos-profiles.yaml
RandyLevensalor: @hdamker I've changed this to just use the Generic404 schema.  That will pick up the additional descriptions after both PRs are merged.

hdamker:
> @hdamker I believe that your comments have been resolved.

@RandyLevensalor Looks good to me. 

BTW: in #326 I also replaced now all "...Notfound404" with the "Generic404" as recommend by @jlurien in https://github.com/camaraproject/QualityOnDemand/pull/326#discussion_r1696760385. All references were in lines you haven't touched in your PR, so I hope it will merge together without conflicts 🤞

hdamker (approved):
LGTM

RandyLevensalor:
@camaraproject/quality-on-demand_codeowners are we ready to merge this?

hdamker:
@RandyLevensalor: I would like to merge first #326 ... in best case there are no conflicts but if yes, the a rebase might be needed.

RandyLevensalor:
@hdamker it doesn't show any merge conflicts

hdamker:
> @hdamker it doesn't show any merge conflicts

@RandyLevensalor Yes, not merge conflict, but unfortunately I've missed to integrate the changes from https://github.com/camaraproject/QualityOnDemand/pull/326/commits/55c609a4a07735654f6c6e985ccac92156fc584d also proactive into qos-profiles.yaml. So either you rebase and apply the changes in your or we need to create a new PR for the following changes afterwards:

* Split of Generic404 into Generic404 (with only "NOT_FOUND") and GenericDevice404 (with "NOT_FOUND" and "DEVICE_NOT_FOUND")
* within the responses the POST endpoint with device: changing "Generic404" to "GenericDevice404", and adding 422 and 429 as valid responses (Generic422 and Generic429 are already there)

hdamker (changes requested):
See https://github.com/camaraproject/QualityOnDemand/pull/318#issuecomment-2269393391 for needed changes

RandyLevensalor:
@hdamker Thanks for the feedback.  Rebased, updated the device and added 422 and 429 as well.

hdamker:
> @hdamker Thanks for the feedback. Rebased, updated the device and added 422 and 429 as well.

Perfect, almost done then ... just delete lines 609 - 614 from `Generic404` (otherwise are `Generic404` and `GenericDevice404` identical 😃).

code/API_definitions/qos-profiles.yaml
```diff
@@ -77,33 +77,25 @@ tags:
 
 paths:
   /qos-profiles:
-    get:
+    post:
       tags:
         - QoS Profiles
-      summary: "Get All QoS Profiles"
+      summary: Get QoS profiles avaialable for a given device
```
hdamker, line 83:
```suggestion
      summary: Get QoS profiles available for a given device
```

hdamker (approved):
LGTM

One type found.

hdamker (approved):

jlurien (approved):

PR #325, opened 2024-07-19
List sessions endpoint for quality-on-demand

### LINKED ISSUE #101: List endpoint for active sessions of authenticated user

jlurien:
#### What type of PR is this?

* enhancement/feature

Adds a new endpoint to return a list of sessions for a given device, thus it will be possible to look for sessions for a given device without knowing the assigned sessionId.

Fixes #101 

Deprecates PR #228

I have made device required in the input, to be aligned with the rest of the API in this PR. There is a separate issue #313 to decide on the alignment of the API with the guidelines regarding device identification and making it optional in the request body.

#### Changelog input

```
* New operation retrieveSessions to get a list of sessions for a given device
```

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -487,7 +487,7 @@ paths:
         "403":
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/Generic404"
+          $ref: "#/components/responses/GenericDevice404"
         "422":
           $ref: "#/components/responses/Generic422"
         "429":
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -445,14 +445,16 @@ paths:
     post:
       tags:
         - QoS Sessions
-      summary: Get all QoS session information for a given device
+      summary: Get QoS session information for a device
       description: |
-        Querying for QoS session resource information details for a given device
+        Querying for QoS session resource information details for a device
 
         **NOTES:**
-        - The access token may be either 2-legged or 3-legged. If a 3-legged access token is used, the end user associated with the session must also be associated with the access token.
-        - If a 2-legged access token is used, all returned sessions must have been created by the same API client given in the access token.
-        - If no QoS session is found for the device, an empty array is returned.
+        - The access token may be either 2-legged or 3-legged.
+          - If a 3-legged access token is used, the end user (and device) associated with the session must also be associated with the access token. In this case it is recommended NOT to include the `device` parameter in the request (see "Handling of device information" within the API description for details).
+          - If a 2-legged access token is used, the device parameter must be provided and identify a device.
+        - The session must have been created by the same API client given in the access token
+        - If no QoS session is found for the requested device, an empty array is returned.
       operationId: retrieveSessions
       parameters:
         - $ref: "#/components/parameters/x-correlator"
@@ -474,8 +476,10 @@ paths:
               schema:
                 $ref: "#/components/schemas/RetrieveSessionsOutput"
               examples:
-                PROVISIONING_AVAILABLE:
+                RETRIEVE_SESSIONS_ONE_ITEM:
                   $ref: "#/components/examples/RETRIEVE_SESSIONS_EXAMPLE"
+                RETRIEVE_SESSIONS_NO_ITEMS:
+                  $ref: "#/components/examples/RETRIEVE_SESSIONS_EMPTY_EXAMPLE"
         "400":
           $ref: "#/components/responses/Generic400"
         "401":
@@ -486,6 +490,8 @@ paths:
           $ref: "#/components/responses/Generic404"
         "422":
           $ref: "#/components/responses/Generic422"
+        "429":
+          $ref: "#/components/responses/Generic429"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
@@ -927,8 +933,6 @@ components:
     RetrieveSessionsInput:
       description: Parameters to get QoS session information by device
       type: object
-      required:
-        - device
       properties:
         device:
           $ref: "#/components/schemas/Device"
@@ -1184,3 +1188,8 @@ components:
           startedAt: "2024-06-01T12:00:00Z"
           expiresAt: "2024-06-01T13:00:00Z"
           qosStatus: AVAILABLE
+
+    RETRIEVE_SESSIONS_EMPTY_EXAMPLE:
+      summary: No sessions found for the device
+      description: An empty array is returned when no sessions are found for the device
+      value: []
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -442,7 +442,7 @@ paths:
         "503":
           $ref: "#/components/responses/Generic503"
 
-  /sessions/retrieve:
+  /retrieve-sessions:
     post:
       tags:
         - QoS Sessions
@@ -452,6 +452,7 @@ paths:
 
         **NOTES:**
         - The access token may be either 2-legged or 3-legged. If a 3-legged access token is used, the end user associated with the session must also be associated with the access token.
+        - If a 2-legged access token is used, all returned sessions must have been created by the same API client given in the access token.
         - If no QoS session is found for the device, an empty array is returned.
       operationId: retrieveSessions
       parameters:
@@ -492,7 +493,7 @@ paths:
           $ref: "#/components/responses/Generic503"
       security:
         - openId:
-            - "quality-on-demand:sessions:read"
+            - "quality-on-demand:sessions:retrieve-by-device"
 
 components:
   securitySchemes:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -464,7 +464,6 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/RetrieveSessionsInput"
-        required: true
       responses:
         "200":
           description: Returns the QoS sessions information for a given device. A device may have multiple sessions, thus the response is an array. An empty array is returned if no sessions are found.
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -464,6 +464,7 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/RetrieveSessionsInput"
+        required: true
       responses:
         "200":
           description: Returns the QoS sessions information for a given device. A device may have multiple sessions, thus the response is an array. An empty array is returned if no sessions are found.
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -981,6 +981,20 @@ components:
             code: PERMISSION_DENIED
             message: "Operation not allowed: ..."
 
+    SessionNotFound404:
+      description: Session not found
+      headers:
+        x-correlator:
+          $ref: '#/components/headers/x-correlator'
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 404
+            code: NOT_FOUND
+            message: "Session Id does not exist"
+
     Generic404:
       description: Not found
       headers:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -1013,38 +1013,6 @@ components:
                 code: DEVICE_NOT_FOUND
                 message: Device identifier not found.
 
-    Generic404:
-      description: Not found
-      headers:
-        x-correlator:
-          $ref: "#/components/headers/x-correlator"
-      content:
-        application/json:
-          schema:
-            $ref: "#/components/schemas/ErrorInfo"
-          examples:
-            GENERIC_404_NOT_FOUND:
-              summary: Generic Not Found
-              description: Resource is not found
-              value:
-                status: 404
-                code: NOT_FOUND
-                message: "{{resource}} is not found"
-
-    Generic422:
-      description: Unprocessable entity
-      headers:
-        x-correlator:
-          $ref: '#/components/headers/x-correlator'
-      content:
-        application/json:
-          schema:
-            $ref: "#/components/schemas/ErrorInfo"
-          example:
-            status: 422
-            code: UNPROCESSABLE_ENTITY
-            message: "Value not acceptable: ..."
-
     SessionInConflict409:
       description: Conflict
       headers:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -455,7 +455,7 @@ paths:
           - If a 2-legged access token is used, the device parameter must be provided and identify a device.
         - The session must have been created by the same API client given in the access token
         - If no QoS session is found for the requested device, an empty array is returned.
-      operationId: retrieveSessions
+      operationId: retrieveSessionsByDevice
       parameters:
         - $ref: "#/components/parameters/x-correlator"
       requestBody:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -312,7 +312,7 @@ paths:
         "403":
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/SessionNotFound404"
+          $ref: "#/components/responses/Generic404"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
@@ -358,7 +358,7 @@ paths:
         "403":
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/SessionNotFound404"
+          $ref: "#/components/responses/Generic404"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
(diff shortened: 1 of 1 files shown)


code/API_definitions/quality-on-demand.yaml
```diff
@@ -436,12 +436,64 @@ paths:
...
+    post:
+      tags:
+        - QoS Sessions
+      summary: Get all QoS session information for a given device
+      description: |
+        Querying for QoS session resource information details for a given device
+
+        **NOTES:**
+        - The access token may be either 2-legged or 3-legged. If a 3-legged access token is used, the end user associated with the session must also be associated with the access token.
+        - If no QoS session is found for the device, an empty array is returned.
```
hdamker, line 455: I think we have to distinct the cases of 2-legged and 3-legged token even more:
- with 2-legged token the API Client can get only the sessions created by the same API Client
- with 3-legged token all session created for the device associated with the access token are returned (at least that was my understanding of a use case at Telefonica?)

As we will make device optional as parameter: should there be a valid return if 2-legged token is used and no device is given? As far as I remember we don't want to return the complete list of sessions created by the API Client, even if this was the original intention of #101. One reason is that we then need to introduce paging.

The text for the two lines could get changed as following:
```suggestion
        - The access token may be either 2-legged or 3-legged. 
          - If a 3-legged access token is used, the end user associated with the session must also be associated with the access token. All sessions created for the given device will be returned
          - If a 2-legged access token is used, all returned sessions must be created by the API client given in the access token. In this case the device parameter must be used and specify a device.
        - If no QoS session is found for the device, an empty array is returned.
```

code/API_definitions/quality-on-demand.yaml
```diff
@@ -872,6 +924,22 @@ components:
         - code
         - message
 
+    RetrieveSessionsInput:
+      description: Parameters to get QoS session information by device
+      type: object
+      required:
+        - device
```
hdamker, line 931: device should be optional, not required

code/API_definitions/quality-on-demand.yaml
```diff
@@ -1024,3 +1110,20 @@ components:
           sessionId: '123e4567-e89b-12d3-a456-426614174000'
           qosStatus: 'UNAVAILABLE'
           statusInfo: 'DURATION_EXPIRED'
+
+    RETRIEVE_SESSIONS_EXAMPLE:
+      summary: List of QoS sessions for the device
+      description: A single QoS session for the device is available
+      value:
```
hdamker, line 1117: Shouldn't that be an array of items?

code/API_definitions/quality-on-demand.yaml
hdamker: I haven't read the last [comment](https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-2232978314) in #101 before making my comment. Fine for me to keep device in this PR required.

code/API_definitions/quality-on-demand.yaml
hdamker: I haven't read the last https://github.com/camaraproject/QualityOnDemand/issues/101#issuecomment-2232978314 in https://github.com/camaraproject/QualityOnDemand/issues/101 before making my comment. Fine for me to postpone this discussion into the context of #313 and #326.

jlurien: Relation with client_id is something that we may have to consider also for the rest of endpoints accessing created sessions, even if we have pragmatically avoided the topic relying on the knowledge of the sessionId as a secret. Case is: 

client_id_A creates a session, client_id_B, with a valid token, tries to retrieve or delete that session. Should we return success or an error? Should we have a different behaviour if input is sessionId or device?

Returning an error, implies that the "owner" client_id has to be kept as a property of the session by the implementation, and only owner client_id can access or operate over a created session. I think that now we are not considering this requirement, but we may need to make explicit the expected behaviour for each scenario, for example to consider the scenario in a test plan. 

There are several scenarios to decide upon:

| Case | Type of token | input | same client_id | Expected behaviour |
| :---: | :---: | :---: | :---: | :--- |
| 1 | 3-legged | session_id | Y | 200 + sessionInfo |
| 2 | 3-legged | session_id | N | 200 + sessionInfo, new client_id needs permission from user |
| 3 | 2-legged | session_id | Y | 200 + sessionInfo, but after checking client_id ownership ? |
| 4 | 2-legged | session_id | N | Now: 200 + sessionInfo, but should we check client_id ownership and return 403?  |
| 5 | 3-legged | device | Y  | 200 + sessionInfo, checking that device is coherent with token |
| 6 | 3-legged | device | N | 200 + sessionInfo, checking that device is coherent with token, new client_id needs permission from user |
| 7 | 2-legged | device | Y | TBD: 200 + sessionInfo, after checking client_id ownership ? |
| 8 | 2-legged | device | N | TBD: 403 after after checking client_id ownership or return session without any check on client_id ownership ? |
| 9 | 3-legged | empty (device deducted from token) | Y | As 5 |
| 10 | 3-legged | empty (device deducted from token) | N | As 6 |
| 11 | 3-legged | empty (device not deducted from token) | Y | 403 INVALID_CONTEXT |
| 12 | 3-legged | empty (device not deducted from token) | N | 403 INVALID_CONTEXT |
| 13 | 2-legged | empty | Y | TDB: Return all sessions for the client_id, or avoid |
| 14  | 2-legged | empty | N | TBD: To be avoided (return all sessions for any client) |

Even if cases 9 and 10 are clear, I tend to think that we should make the device required by now, to avoid the other corner cases (13 and 14). And as as commented in #313 making device optional has other implications.

I have doubts with cases 4 and 7/8, after what I wrote above about implications on client_id ownership: should we keep a different behaviour depending on the type of input or have a common approach (checking or not checking ownership at all)? I tend to think here that we should have the same guideline for scenarios 4 and 8.

It may add flexibility to distinguish the endpoint (retrieve session by device) with a different scope, so providers can restrict the usage of 2-legged tokens for certain apps or to certain endpoints.

code/API_definitions/quality-on-demand.yaml
jlurien: It is an array of items indeed, but with only one element

code/API_definitions/quality-on-demand.yaml
hdamker: If we keep the scope of the endpoint to "retrieve session by device", meaning that there need to be an identified device, then we can still make the parameter device optional, as for the cases 13 and 14 there is a clear behaviour defined in the appendix of the Design Guidelines (which I copied as required into the info.description):

> ### Error handling for unidentifiable devices:
> - If the `device` object is not included in the request and the device information cannot be derived from the 3-legged access token, the server will return a 422 `UNIDENTIFIABLE_DEVICE` error.
> ### Restrictions for tokens without an associated authenticated identifier:
> - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.

BTW: within 11 and 12 the error code would be UNIDENTIFIABLE_DEVICE. INVALID_TOKEN_CONTEXT can happen in 5, 6, 7 and 8. And I'm in favour to check the ownership of a session, I don't see cases where a client_id_B should manipulate the session of a client_id_A.

jlurien: > And I'm in favour to check the ownership of a session, I don't see cases where a client_id_B should manipulate the session of a client_id_A.

A use case that we have for this is when the user uses an App to manage or know about the existence of sessions that may have been created from another channel. But 3-legged tokens are required as the App must have consent from the user to manage their sessions. What I don't foresee is a use case where an App using 2-legged tokens can manage sessions created by other apps for any user.

hdamker: > A use case that we have for this is when the user uses an App to manage or know about the existence of sessions that may have been created from another channel. But 3-legged tokens are required as the App must have consent from the user to manage their sessions. What I don't foresee is a use case where an App using 2-legged tokens can manage sessions created by other apps for any user.

In opposite to my initial comment: I don't think we can cover this use case with the same endpoint. You have defined (correctly, I would say) as scope for the endpoint `quality-on-demand:sessions:read`. There is no way for the authorization server to decide if the client_id_A want to read only the own sessions (covered by "read") or want to have the right to get all sessions of the device listed. That must then be requested with a different scope, some "sudo / manage" on the device, so that the device user can be asked for consent to this scope.

eric-murray: The question in my mind is - if a given end user has consented to `quality-on-demand:sessions:read` scope for both Client A and Client B, do they intend that Client A has read access to sessions created by Client B (and vice-versa)?

In general, I would think not. Hence only sessions created by the authenticated client for the identified end user (whether via a 3-legged token or explicitly specified) should be returned. Even if a 2-legged token is used, implying no end user consent is required, I still think sessions created by Client B should not be read by Client A.

If it is desired to know all existing sessions for a given end user irrespective of client, I would say the solution is that this new endpoint should be called multiple times - once for each client that is authorised to create sessions for that end user. Maybe we would like an "admin client" to have access to everything, but they should not use the public interface for that.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -436,12 +436,64 @@ paths:
...
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/SessionNotFound404"
+          $ref: "#/components/responses/Generic404"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
           $ref: "#/components/responses/Generic503"
 
+  /sessions/retrieve:
```
eric-murray, line 445: I also raised this point for the QoS provisioning proposal, but the design guidelines could be interpreted as meaning this endpoint must be named `POST /retrieve-sessions`.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -872,6 +924,22 @@ components:
...
         - message
 
+    RetrieveSessionsInput:
+      description: Parameters to get QoS session information by device
+      type: object
+      required:
+        - device
+      properties:
+        device:
+          $ref: "#/components/schemas/Device"
```
eric-murray, line 934: If a session was created using, for example, IP address / port, should we be allowed to search for it by other parameters, such as MSISDN? Would we just leave this up to implementations?

code/API_definitions/quality-on-demand.yaml
jlurien: > The question in my mind is - if a given end user has consented to `quality-on-demand:sessions:read` scope for both Client A and Client B, do they intend that Client A has read access to sessions created by Client B (and vice-versa)?
> 
> In general, I would think not. Hence only sessions created by the authenticated client for the identified end user (whether via a 3-legged token or explicitly specified) should be returned. Even if a 2-legged token is used, implying no end user consent is required, I still think sessions created by Client B should not be read by Client A.
> 
> If it is desired to know all existing sessions for a given end user irrespective of client, I would say the solution is that this new endpoint should be called multiple times - once for each client that is authorised to create sessions for that end user. Maybe we would like an "admin client" to have access to everything, but they should not use the public interface for that.

That is indeed the question, what "quality-on-demand:sessions:read" scope means. If we explicitly indicate that the scope give access to a client to read sessions **only created by that client**, only clients that created sessions can get or delete those sessions. It is also possible to create different scopes  "quality-on-demand:sessions:read:self" and "quality-on-demand:sessions:read:all" to differentiate the behaviour and only assign the later to privileged apps.

Ultimately we have to define a test scenario such as:

```
Given a clientId_A created a session with sessionId=X
And clientId_B has an an access token with scope "quality-on-demand:sessions:read"  
When clientId_B requests the session for sessionId=X
Then the response is ??
```
Implications on implementation are not trivial, as Auth servers granting and checking scopes will not generally know anything about sessionId and owner of that session. It would have to be the service implementation the one enforcing the rule and for that it needs to attach the clientId during session creation. But clientId is not part of the API, it is part of the token, so info contained in the access token has to be populated from the Auth server to the service implementation. This affects any API dealing with resources, such as subcriptions. So we probably need an explicit guideline.

For the short term, we have to take a decision regarding this endpoint and its scopes.

code/API_definitions/quality-on-demand.yaml
jlurien: I also commented there, that both names work for me, so no strong opinion here

code/API_definitions/quality-on-demand.yaml
jlurien: It is hard to limit that in an API. It depends on how implementations attach device info to the session and are able to work with or exchange different type of identifiers. They may locate the session or not depending on this.

But the concept of Device as property of current SessionInfo triggers many questions that we probably need to tackle for future releases. So far we ask the device to be returned as part of sessionInfo in responses, but we have not stated explicitly if the identifiers in the response must always be the same identifiers that were indicated in the creation request. Moreover, IP addresses may be reassigned and they may no longer identify the same device at all moments. Also, if we decide to map to a permanent identifier, such a MSISDN, and we return it, we may be disclosing private and unnecessary information to the client.

code/API_definitions/quality-on-demand.yaml
hdamker: My proposal would be to be restrictive here: 

- We require that the returned sessions are
  - associated with the client_id presented in the access token
  - in case of 3-legged tokens the end user (or device) associated with the session must also be associated with the access token
- We require that a device must be identifiable for the operation (read: we don't list all sessions associated with a client_id)
- We define an own scope for the operation, not using `read`

With that we would support only the cases 5,7,9 from the table above for the retrieve operation (session is not an input option here). Operators can decide separately from the `read` scope if they grant access.

And I suppose my suggested text from the begin of the discussion thread might fit to these restrictions:

>  - The access token may be either 2-legged or 3-legged. 
>    - If a 3-legged access token is used, the end user associated with the session must also be associated with the access token. In this case it is recommended to not use the device parameter (see API description for details).
>    - If a 2-legged access token is used, all returned sessions must be created by the API client given in the access token. In this case the device parameter must be used and specify a device.
> - If no QoS session is found for the device, an empty array is returned.

jlurien: The proposal makes sense to me. I think it is reasonable to restrict the current scope "read" only to the clientId that created the sessions. We would have to make that more explicit for the whole API.

jlurien: I will not add by now any reference to the optionality of device, as that is not specific to this PR and we should have consistent texts for that across all endpoints, so I'll just add:

        - If a 2-legged access token is used, all returned sessions must have been created by the same API client given in the access token.

code/API_definitions/quality-on-demand.yaml
jlurien: I have changed it to /retrieve-sessions

code/API_definitions/quality-on-demand.yaml
hdamker: The requirement that "all returned sessions must have been created by the same API client given in the access token" is independent of 2- or 3-legged token. So also my text proposal was wrong:

- The access token may be either 2-legged or 3-legged.
  - all returned sessions must have been created by the same API client given in the access token
  - If a 3-legged access token is used, the end user associated with the session must also be associated with the access token. In this case it is recommended to not use the device parameter (see API description for details).
  - If a 2-legged access token is used the device parameter must be used and specify a device.
- If no QoS session is found for the device, an empty array is returned.

hdamker:
We need to consider if adding of the operation with the restrictions defined in the above discussion thread (see this comment as summary) does still make sense.

The original issue #101 asked for "A /get endpoint offering a paginated response of all sessions that the currently authenticated user created and are still active." This we excluded, it would a different operation if offered at all.

Then there was the idea of an app on the device which would allow the device user to see all active QoD session, independent of who created them. This we excluded as well. I would argue that this would be an operation which needs to be very carefully designed and potentially not provided beyond operator owned apps.

If these two use cases are not covered by the current proposal ... what value is left for developers with this new operation?
We need to balance that with the effort to implement and the privacy risks which we identified during the discussion (e.g. the question raised by https://github.com/camaraproject/QualityOnDemand/pull/325#discussion_r1690551724)

jlurien:
I (hopefully) have adapted the new endpoint to the same principles in #326

code/API_definitions/quality-on-demand.yaml
```diff
@@ -441,6 +441,65 @@ paths:
...
+      description: |
+        Querying for QoS session resource information details for a device
+
+        **NOTES:**
+        - The access token may be either 2-legged or 3-legged.
+          - If a 3-legged access token is used, the end user (and device) associated with the session must also be associated with the access token. In this case it is recommended NOT to include the `device` parameter in the request (see "Handling of device information" within the API description for details).
+          - If a 2-legged access token is used, the device parameter must be provided and identify a device.
+        - The session must have been created by the same API client given in the access token
+        - If no QoS session is found for the requested device, an empty array is returned.
+      operationId: retrieveSessions
```
jlurien, line 458:
```suggestion
      operationId: retrieveSessionsByDevice
```

code/API_definitions/quality-on-demand.yaml
```diff
@@ -441,6 +441,65 @@ paths:
...
+      operationId: retrieveSessionsByDevice
+      parameters:
+        - $ref: "#/components/parameters/x-correlator"
+      requestBody:
+        description: Parameters to get QoS session information by device
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/RetrieveSessionsInput"
+        required: true
```
jlurien, line 467:
```suggestion
```

code/API_definitions/quality-on-demand.yaml
```diff
@@ -464,7 +464,6 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/RetrieveSessionsInput"
-        required: true
```
hdamker: There is a discussion about this within https://github.com/camaraproject/Commonalities/issues/247. What I'm getting from the discussion that the preferred option seems to make the structure not optional, even if it can be empty, see e.g. https://github.com/camaraproject/Commonalities/issues/247#issuecomment-2242341834

jlurien: I had my doubts about this but looking into the discussion I will revert it to required: true

code/API_definitions/quality-on-demand.yaml
```diff
@@ -441,6 +441,64 @@ paths:
...
+      operationId: retrieveSessionsByDevice
+      parameters:
+        - $ref: "#/components/parameters/x-correlator"
+      requestBody:
+        description: Parameters to get QoS session information by device
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/RetrieveSessionsInput"
+      responses:
```
jlurien, line 467:
```suggestion
        required: true
      responses:
```

code/API_definitions/quality-on-demand.yaml
```diff
@@ -441,6 +441,65 @@ paths:
...
+                RETRIEVE_SESSIONS_NO_ITEMS:
+                  $ref: "#/components/examples/RETRIEVE_SESSIONS_EMPTY_EXAMPLE"
+        "400":
+          $ref: "#/components/responses/Generic400"
+        "401":
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/Generic403"
+        "404":
+          $ref: "#/components/responses/Generic404"
```
hdamker, line 490: #326 has now GenericDevice404, so you can change to it here:

```suggestion
        "404":
          $ref: "#/components/responses/GenericDevice404"
```

jlurien:
I should be ready to be merged after #326

RandyLevensalor:
> I should be ready to be merged after #326

@jlurien #326 has been merged.  So you'll need to rebase this branch to pick up those changes and hopefully clear the linting error.

hdamker (approved):

hdamker:
@eric-murray @RandyLevensalor also here the question if you have any additional comments?

jlurien:
@hdamker please feel free to merge it once you get all needed OKs

RandyLevensalor (approved):
LGTM

PR #331, opened 2024-07-26
Add jlurien as  CODEOWNER

jlurien:
#### What type of PR is this?

* subproject management

Add new codeowner


hdamker (approved):
Was agreed within QoD call on July 26th.
Added @jlurien to @camaraproject/quality-on-demand_codeowners and checked codeowner mailing list.

PR #344, opened 2024-08-12
Quality on Demand: Fix RETRIEVE_SESSIONS_EXAMPLE

### LINKED ISSUE #343: Quality on Demand example RETRIEVE_SESSIONS_EXAMPLE must use sink
**Problem description**
It mentions legacy webhook.notificationUrl

**Expected behavior**
Change example to sink

**Alternative solution**

**Additional context**
To be fixed for rc.2, it can be accumulated with future changes.

jlurien:
#### What type of PR is this?

* bug

- Fix example which was not compliant with new schema
- Main version set to wip again

Fixes #343 

To be considered for r1.2 (RC2)

#### Changelog input

```
- Fix RETRIEVE_SESSIONS_EXAMPLE
```

### DIFF
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -60,7 +60,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0-rc.1
+  version: wip
   x-camara-commonalities: 0.4.0-rc.1
 
 externalDocs:
@@ -68,7 +68,7 @@ externalDocs:
   url: https://github.com/camaraproject/QualityOnDemand
 
 servers:
-  - url: "{apiRoot}/qod-provisioning/v0.1rc1"
+  - url: "{apiRoot}/qod-provisioning/vwip"
     variables:
       apiRoot:
         default: http://localhost:9091
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -55,7 +55,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.11.0-rc.1
+  version: wip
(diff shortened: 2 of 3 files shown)


code/API_definitions/qod-provisioning.yaml
```diff
@@ -60,15 +60,15 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0-rc.1
+  version: wip
```
eric-murray, line 63: If we want to fix this for the meta-release, shouldn't this be `version: 0.1.0-rc.2`, rather than going back to `wip` at this stage?

Same comment for server URL, and for other YAMLs.

hdamker: @eric-murray going back to wip is correct ... it is indicating that these commits are not part of a release, but the work in progress between the releases. There could be further bug fixes to the same file before we create the next (pre-)release. The version will be set again in a release PR ... could be either 0.1.0-rc.2 (if it would make still make sense) or already 0.1.0.

eric-murray: OK. In that case, I think I'd prefer a specific PR to reset `version` and `url` immediately after a release, independent of any other changes. Otherwise all subsequent PRs need to include these changes until one of them is merged.

For this PR, we need to decide if this change will be part of the next release or not before merging.

jlurien: The problem would be if we wanted to work in parallel: fixes for r1.2 (RC.2) and new content for a future meta-release (r2.1). Without branches that is not possible.

hdamker: For this fix of examples I guess the decision is clear that we want to have them in r1.2. Therefore I'm in favour to merge it asap (to resolve also the version reset).

hdamker (approved):

eric-murray (approved):

RandyLevensalor (approved):
LGTM

PR #356, opened 2024-08-30
Add new error schema to extendDuration: SessionStateInConflict409 with QUALITY_ON_DEMAND.SESSION_EXTENSION_NOT_ALLOWED 

### LINKED ISSUE #351: Behaviour when extending not AVAILABLE session
**Problem description**
Current spec do not indicate the expected behaviour when the extension of the duration of a session in status other than AVAILABLE is requested

**Expected action**
Agree on the expected result, and update spec and tests. 

For sessions with qosStatus = REQUESTED, implementation may accept the request and take into account the petition when the session is activated, or return a success without modifying the original duration (ignoring the request), or even return an error indicating that this request is not acceptable  in  the current state. 
For sessions with qosStatus = UNAVAILABLE, implementation may return a success without modifying the duration (ignoring the request), or return an error indicating that this request is not acceptable  in the current state.

hdamker:
#### What type of PR is this?

* correction

Clarifies the undefined behaviour if API consumer tries to extend a session which is state REQUESTED or UNAVAILABLE as discussed in #351 

Fixes #351

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -1186,7 +1186,7 @@ components:
         application/json:
           schema:
             $ref: "#/components/schemas/ErrorInfo"
-          example:
+          examples:
             SessionExtensionNotAllowed:
               description: Session extension is in conflict with current session status
               value:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -336,7 +336,7 @@ paths:
         - QoS Sessions
       summary: "Extend the duration of an active session"
       description: |
-        Extend the overall session duration of an active QoS session.
+        Extend the overall session duration of an active QoS session (with qosStatus = AVAILABLE).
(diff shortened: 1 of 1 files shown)


jlurien (approved):
LGTM

code/API_definitions/quality-on-demand.yaml
```diff
@@ -1175,6 +1177,23 @@ components:
...
+    SessionStatusConflict409:
+      description: Conflict
+      headers:
+        x-correlator:
+          $ref: '#/components/headers/x-correlator'
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
```
jlurien, line 1189:
```suggestion
          examples:
```

jlurien (approved):
LGTM

hdamker:
@eric-murray @RandyLevensalor please have as well a look on this last minute change - should be ok, but more eyes might see more.

RandyLevensalor (approved):

PR #364, opened 2024-09-25
Patch wiki links

hdamker:
#### What type of PR is this?

* correction
* subproject management

Correct wiki links after the migration (former short links are no longer working within the new wiki space).

Fixes # na


RandyLevensalor (approved):
LGTM

Issue #302, opened 2024-06-13
Providing developers with alternate to QoS profile


maheshc01:
**Problem description**
Today QoD API requires the application developer to pass in qosProfile as one of the mandatory params for creating POST  /sessions. as described in the yaml this qos profile can vary from each operators which puts the burden on the developers to keep themselves updated on all the qos profiles supported by each operators across the world. 
I hope we all can see the problem with this. It goes against the basic CAMARA principle of requiring the APIs to be developer friendly.

**Solution**
Rather than the developer having to learn all the qos profile options across the different operators and figure out which suits best for their application it should be the operator's responsibility to figure out which is the best qos profile they need to apply based on the application needs as well as additional context based on user subscription data.
Proposal is to adopt the concept of application profiles as defined in connectivity insights (which is already modeled to work closely with QoD API) via which the application developers define what they expect in terms of network performance and which calling the QoD /sessions API, they can pass on application profile id as an alternate to qos profile and the operator will take on the responsibility to identify the best qos profile and create the session based on that.

**Additional context**
there is an open issue in commonalities on the concept of application Id. Discussions here and connectivity insights can be used as references to be adopted in commonalities.
another advantage with this proposal is that it puts us on the path of decoupling QCI based implantation to support that API request and in further the same spec can be used to provide network slice based quality on demand.

jlurien:
Hi @maheshc01,

The topic is interesting. Looking also into the PR, it seems like the proposal is about giving developers/applications the ability to create their own adhoc profiles, more than figuring out which of the available QoS Profiles are closer to their needs. I'm not very familiar with [Connectivity Insights API](https://github.com/camaraproject/ConnectivityInsights/blob/main/code/API_definitions/CAMARA%20Connectivity%20Insights%20API.yaml), but I see that they model a `policyId` instead of the `applicationProfileId` you are proposing in the PR, are they equivalent?

To me they are **2 different requirements**:

* **To figure out what is the best available QoS Profile, among the available ones**. For this, we could enhance the qos-profiles API with some complex filters based on the profiles characteristics. But that only would make sense if implementations had lots of profiles with detailed characteristics. 

* **To allow applications to create adhoc profiles which some required characteristics**. For this, the complex part would be in the negotiation phase, when the implementation has to assure that the request can be fulfilled, which probably has to take into account many variables, such as device, requested characteristics, time interval, etc.. Also it is not clear how developers would know and negotiate/accept concepts like pricing for the request, SLAs, etc

[...]

PR #378, opened 2024-11-26
Add guidance for multi-SIM scenarios

### LINKED ISSUE #376: Document current behavior for devices with multiple SIMs
**Problem description**
Until #362 is resolved, it is unclear how the API will support devices with multiple SIMs with the MSIDN is used as the device identifier.

**Expected action**
The expected behavior and possible error conditions along with suggestions on how to avoid these errors should be added to the API documentation.

**Additional context**

eric-murray:
#### What type of PR is this?
* documentation

Provides guidance on using the QoD APIs in multi-SIM scenarios

Fixes #376 

This guidance will need to be updated when Commonalities Issue [#302](https://github.com/camaraproject/Commonalities/issues/302) is resolved

#### Changelog input

```
 release-note
 - Add guidance for multi-SIM scenarios
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -65,7 +65,7 @@ info:
     Possible solutions in such a scenario include:
     - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device
     - Identifying the intended device from a unique identifier for that device, such as its source IP address and port
-    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device, and use the secondary phone number to identify the intended device
+    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device, and use the secondary phone number to identify the intended device if available.
 
   license:
     name: Apache 2.0
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -60,7 +60,7 @@ info:
 
     # Multi-SIM scenario handling
 
-    In multi-SIM scenarions where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
 
     Possible workarounds in such a scenario include:
     - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -57,6 +57,16 @@ info:
     ### Restrictions for tokens without an associated authenticated identifier:
 
     - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+    # Multi-SIM scenario handling
+
+    In multi-SIM scenarions where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
+
+    Possible workarounds in such a scenario include:
+    - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device
+    - Identifying the intended device from a unique identifier for that device, such as its source IP address and port
+    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device, and use the secondary phone number to identify the intended device
+
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -60,9 +60,9 @@ info:
 
     # Multi-SIM scenario handling
 
-    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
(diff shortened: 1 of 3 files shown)


code/API_definitions/qod-provisioning.yaml
```diff
@@ -57,6 +57,16 @@ info:
     ### Restrictions for tokens without an associated authenticated identifier:
 
     - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+    # Multi-SIM scenario handling
+
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
+
+    Possible workarounds in such a scenario include:
```
jlurien, line 65:
```suggestion
    Possible mitigation in such a scenario include:
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -57,6 +57,16 @@ info:
...
 
     - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+    # Multi-SIM scenario handling
+
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
+
+    Possible workarounds in such a scenario include:
+    - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device
+    - Identifying the intended device from a unique identifier for that device, such as its source IP address and port
```
jlurien, line 67:
```suggestion
    - Identifying the intended device from a unique identifier for that device, such as its source IP address and port. This applies as well as the recommended value for the `login_hint` parameter of the OIDC CIBA flow.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -57,6 +57,16 @@ info:
...
+    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device, and use the secondary phone number to identify the intended device
```
jlurien, line 68:
```suggestion
    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device, and use the secondary phone number to identify the intended device if available.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -57,6 +57,16 @@ info:
     ### Restrictions for tokens without an associated authenticated identifier:
 
     - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+    # Multi-SIM scenario handling
+
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
+
+    Possible workarounds in such a scenario include:
+    - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device
```
jlurien, line 66: If we finally approve in Commonalities that a device identifier in request body is incompatible with a 3-legged access token (and "In this case, a `422 UNNECESSARY_IDENTIFIER` error code MUST be returned"), then we can remark that 
```suggestion
    - Using the authorisation code flow to obtain an access token, which will automatically identify the intended device. Moreover, in this case, no other device identifier must be included in the request.
```

jlurien (commented):
In line with the latest consensus in https://github.com/camaraproject/Commonalities/pull/324, I have included some suggestions. They will apply to all the 3 APIs with minor adjustments.

code/API_definitions/qod-provisioning.yaml
```diff
@@ -57,6 +57,16 @@ info:
     ### Restrictions for tokens without an associated authenticated identifier:
 
     - For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+    # Multi-SIM scenario handling
+
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to a device other than the intended device.
```
jlurien, line 63: Another possibility is to apply the service to **all** the devices associated to the phone number:

```suggestion
    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device to which the QoS provisioning should apply from that phone number. If the phone number is used as the device identifier when provisioning in a multi-SIM scenario, the API may respond with an error, or may apply the provisioning to all the devices associated with a phone number.
```

hdamker: I'm here on the side of @jlurien option: either an error or all the devices associated with the phone number. Applying the provisioning to a device other than the intended device would be an erroneous behavior which the API consumer can not detect. That leaves for an API provider who doesn't want to apply the provisioning to all devices still the option to respond with an error (btw: which one?).

eric-murray: I was thinking of the scenario where an API provider allows the use of secondary MSISDNs for provisioning, but the API consumer mistakenly uses the primary MSISDN when intending to provision one of the secondary devices. Thus provisioning is applied to the primary device (smartphone) when their intention was to provision, say, the tablet device, **but they didn't use the API correctly**.

I think we have two options here:
- Allow the use of secondary MSISDNs for individual device provisioning, with sufficient documentation being provided to the API consumer to avoid them using the wrong MSISDN; or
- Provision for all devices when the primary MSISDN is used, otherwise provision for the secondary device associated with the secondary MSISDN.

The issues with this second solution are:
- the primary device cannot be provisioned individually
- not all devices may support the requested QoS profile
- every implementation will be required to do a multi-SIM check for every API invocation, which is a complication I could well live without for a first commercial version of this API

I appreciate that some implementations may use this second option, so will update the text accordingly. But until the general issue on multi-SIM handling in Commonalities is resolved, I think multi-SIM handling for the QoD APIs should remain somewhat "proprietary".

code/API_definitions/qod-provisioning.yaml
eric-murray: I propose the word "solutions" in the latest update

code/API_definitions/qod-provisioning.yaml
eric-murray: The proposed changes are a patch to the Fall 24 Meta-Release version, where a device identifier is allowed even if a 3-legged token is used. Rather than repeat the rule here, the proposal is to rely on the dedicated section in the `info.description` to explain it.

code/API_definitions/qod-provisioning.yaml
eric-murray: The discussion in the meeting was not to include a specific comment on the use of `ipport` for CIBA, as that anyway follows as an option if the developer is able to use IP/port as the device identifier.

hdamker (approved):

RandyLevensalor (approved):

jlurien (approved):
Thanks @eric-murray, LGTM

PR #384, opened 2024-12-10
Add L4S/DSCP support to QoS Profile

### LINKED ISSUE #173: Add support for DSCP markings for QoD sessions
**Problem description**
A user may want to apply a QoS Profile to a class of traffic based on the DSCP value according to [RFC4594](https://datatracker.ietf.org/doc/html/rfc4594) in the IP header.  This would be traffic from the user device to / from any network and would not necessarily be restricted to a specific application server.

This aligns with traffic classification as defined in the [Home Devices QoD API](https://github.com/camaraproject/HomeDevicesQoD/blob/main/code/API_definitions/home_devices_qod.yaml)

**Possible evolution**
Instead of requiring the application server in the create session schema, add a flow identifier schema that includes application server, DSCP and can contain other attributes which could be used to classify the traffic.

**Alternative solution**

**Additional context**

benhepworth:
#### What type of PR is this?

* enhancement/feature

Add L4S/Diffserv support to Qos Profile

Fixes #173 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -345,7 +345,23 @@ components:
             - non-l4s-queue
             - l4s-queue
             - mixed-queue
-          description: Specifies the type of queue for L4S traffic management
+          description: |
+            Specifies the type of queue for L4S (Low Latency, Low Loss, Scalable Throughput) traffic management. L4S is an advanced queue management approach designed to provide ultra-low latency and high throughput for internet traffic, particularly beneficial for interactive applications such as gaming, video conferencing, and virtual reality.
+
+            **Queue Type Descriptions:**
+
+            - **non-l4s-queue**:
+              A traditional queue used for legacy internet traffic that does not utilize L4S enhancements. It provides standard latency and throughput levels.
+
+            - **l4s-queue**:
+              A dedicated queue optimized for L4S traffic, delivering ultra-low latency, low loss, and scalable throughput to support latency-sensitive applications.
+
+            - **mixed-queue**:
+              A shared queue that can handle both L4S and traditional traffic, offering a balance between ultra-low latency for L4S flows and compatibility with non-L4S flows.
+
+          externalDocs:
+            description: For more details, refer to the L4S standard.
+            url: https://datatracker.ietf.org/doc/rfc9330/
         differentiatedServicesCodepoint:
           type: string
           description: |
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -355,7 +355,7 @@ components:
             example: 3
           minItems: 0
           maxItems: 64
-          description: A list of unique DSCP values from 0-63
+          description: A list of unique DSCP values from 0-63, where each decimal value is mapped to the 6 bit DSCP value. For example `0` would be `000000` and `8` would be `001000` - where 0 and 8 would map to `CS0` and `CS1` respectively as defined in the [IANA DSCP Registry](https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml). This registry defines how [RFC 2474](https://www.rfc-editor.org/rfc/rfc2474) is to be used, while [RFC4594](https://datatracker.ietf.org/doc/html/rfc4594) defines recommended DSCPs to use for different application types. It is worth noting that some operating systems limit what DSCP values can be chosen. For example, Windows defines them for developers [this way](https://learn.microsoft.com/en-us/windows/win32/api/qos2/ne-qos2-qos_traffic_type). Therefore this allows multiple numeric DSCP values to be chosen in order to cover all operating systems and platforms.
       required:
         - name
         - status
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -362,16 +362,19 @@ components:
           externalDocs:
             description: For more details, refer to the L4S standard.
             url: https://datatracker.ietf.org/doc/rfc9330/
-        differentiatedServicesCodepoint:
+        serviceClass:
           type: string
           description: |
-            The name of the service class requested by the API client. It is associated with QoS behaviors optimized for a particular application type.
 
-            There are two sources of DSCP value supported:
-            1. Values that map to the [RFC4594](https://datatracker.ietf.org/doc/html/rfc4594) guidelines
+            The name of the service class requested by the API client. It is associated with QoS behaviors optimized for a particular application type. This aligns with the serviceClass concept used in HomeDevicesQoQ for consistent terminology.
+
+            Service classes define specific QoS behaviors that map to DSCP (Differentiated Services Code Point) values or Microsoft QoS traffic types.
+
+            The supported mappings are:
+            1. Values aligned with the [RFC4594](https://datatracker.ietf.org/doc/html/rfc4594) guidelines for differentiated traffic classes.
             2. Microsoft [QOS_TRAFFIC_TYPE](https://learn.microsoft.com/en-us/windows/win32/api/qos2/ne-qos2-qos_traffic_type) values for Windows developers.
 
-            The following service classes are supported:
+            **Supported Service Classes**:
 
             | Service Class Name    | DSCP Name | DSCP value (decimal) | DCSP value (binary) | Microsoft Value | Application Examples                                                 |
             |-----------------------|-----------|----------------------|---------------------|-----------------|----------------------------------------------------------------------|
@@ -387,15 +390,15 @@ components:
             | Standard              |  DF(CS0)  |           0          |        000000       |        0        | Undifferentiated applications - also:                                |
             |                       |           |                      |                     |                 | Microsoft QOSTrafficTypeBestEffort                                   |
           enum:
-            - microsoft_voice
-            - microsoft_audio_video
-            - real_time_interactive
-            - multimedia_streaming
-            - broadcast_video
-            - low_latency_data
-            - high_throughput_data
-            - low_priority_data
-            - standard
+            - microsoft_voice                # Microsoft QOSTrafficTypeVoice or Control
+            - microsoft_audio_video          # Microsoft QOSTrafficTypeExcellentEffort or AudioVideo
+            - real_time_interactive          # Video conferencing and interactive gaming
+            - multimedia_streaming           # Streaming video or audio on demand
+            - broadcast_video                # Broadcast TV and live events
+            - low_latency_data               # Client/server transactions Web-based ordering
+            - high_throughput_data           # STore and forward applications
+            - low_priority_data              # Low-priority background traffic
+            - standard                       # Default traffic class
           example: real_time_interactive
       required:
         - name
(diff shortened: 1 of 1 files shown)


code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,20 @@ components:
...
+          type: string
+          enum:
+            - non-l4s-queue
+            - l4s-queue
+            - mixed-queue
+          description: Specifies the type of queue for L4S traffic management
+        differentiatedServicesCodepoint:
+          type: array
+          items:
+            type: integer
```
RandyLevensalor, line 352:
```suggestion
            type: integer
            minimum: 0
            maximum: 63
            example: 3
```

code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,20 @@ components:
...
+          minItems: 0
+          maxItems: 64
+          description: A list of unique DSCP values from 0-63
```
RandyLevensalor, line 355:
```suggestion
          description: A list of unique DSCP values from 0-63, where each decimal value is mapped to the 6 bit DSCP value.  For example `0` would be `000000` and `8` would be `001000`.  Where 0 and 8 would map to `CS0` and `CS1` [RFC 2474](https://www.rfc-editor.org/rfc/rfc2474).
```
We should also add the mapping to some alternate standards as well.

RandyLevensalor (changes requested):

code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,23 @@ components:
...
+        differentiatedServicesCodepoint:
+          type: array
+          items:
+            type: integer
+            minimum: 0
+            maximum: 63
+            example: 3
+          minItems: 0
+          maxItems: 64
+          description: A list of unique DSCP values from 0-63
```
benhepworth, line 358:
```suggestion
          minItems: 0
          maxItems: 64
          description: A list of unique DSCP values from 0-63, where each decimal value is mapped to the 6 bit DSCP value. For example `0` would be `000000` and `8` would be `001000` - where 0 and 8 would map to `CS0` and `CS1` respectively as defined in the [IANA DSCP Registry](https://www.iana.org/assignments/dscp-registry/dscp-registry.xhtml). This registry defines how [RFC 2474](https://www.rfc-editor.org/rfc/rfc2474) is to be used, while [RFC4594](https://datatracker.ietf.org/doc/html/rfc4594) defines recommended DSCPs to use for different application types. It is worth noting that some operating systems limit what DSCP values can be chosen. For example, Windows defines them for developers [this way](https://learn.microsoft.com/en-us/windows/win32/api/qos2/ne-qos2-qos_traffic_type). Therefore this allows multiple numeric DSCP values to be chosen in order to cover all operating systems and platforms.
```

benhepworth (commented):
considering your changes, here is some additional verbiage @RandyLevensalor

code/API_definitions/qos-profiles.yaml
benhepworth: I added some additional information - please see my suggestion below

code/API_definitions/qos-profiles.yaml
benhepworth, line 1: @RandyLevensalor @hdamker - I switched from an array to enum to match the QoD Home Profile. I also added the Microsoft DSCP values which some align and some are different. Looking for feedback this week so we can get this into the Spring 25 release.

code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,48 @@ components:
           minimum: 1
           maximum: 10
           example: 3
+        l4sQueueType:
```
hdamker, line 342: Compared to other parameter descriptions is this one too short. At least should the abbreviation l4s be explained (best with a reference to the standard), also the queue types would benefit from a short description.

code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,48 @@ components:
...
+          type: string
+          enum:
+            - non-l4s-queue
+            - l4s-queue
+            - mixed-queue
+          description: Specifies the type of queue for L4S traffic management
+        differentiatedServicesCodepoint:
```
hdamker, line 349: Could we use here as parameter name also "serviceClass", aligned with HomeDevicesQoD?

My thought: the "serviceClass" indicates for which kind of traffic the QoS Profile (as offered by the network) is offered, but DSCP might or might not be used on segments of the e2e connection. With generalization the parameter might get useful also for cases where no DSCP is used within the actual implementation.

code/API_definitions/qos-profiles.yaml
```diff
@@ -339,6 +339,48 @@ components:
...
+          type: string
+          description: |
+            The name of the service class requested by the API client. It is associated with QoS behaviors optimized for a particular application type.
```
hdamker, line 352: The QoS Profile is the description of the quality of service offered by the network when the profile is activated, it is not "requested" by the API Client (the API client can only request the QoS Profile as described).

Some suggestion, but need to be improved:
```suggestion
            The name of a Service Class, indicating that the QoS Profile is offering an optimized behavior for a particular application type. DSCP might or might not be used within the network when the QoS profile is used in a QoS Session (it could be used also only on certain segments of the e2e connection).
```

hdamker (commented):
Some thoughts from my side

RandyLevensalor (dismissed):

benhepworth:
@camaraproject/quality-on-demand_codeowners - this PR should be done, please provide feedback

code/API_definitions/qos-profiles.yaml
hdamker: That one was resolved without comment?

hdamker:
> this PR should be done, please provide feedback

@benhepworth would be great if you can consider my comments above ... I'm not yet happy with the PR in the current form. We should also consider to release the API with another initial version instead of declaring it already stable, as the added parameters are more than a minor change and might need some revision in a next version. 
cc: @eric-murray @jlurien further thoughts?

code/API_definitions/qos-profiles.yaml
benhepworth: thanks for the feedback @hdamker - I updated the description to be more verbose based on your feedback including what L4S is, queue type descriptions, and link to RFC9330.

code/API_definitions/qos-profiles.yaml
benhepworth: I updated to alight with Home Devices QoD - take a look and let me know what you think. Thx @hdamker !

code/API_definitions/qos-profiles.yaml
benhepworth: oops, not sure what happened @hdamker - thanks for the feedback. I updated the description - take a look and let me know if you have any additional feedback.

benhepworth:
ok @hdamker - thanks for all the feedback. I updated everything based on your comments. Let me know if you have any additional feedback. Thanks!

jlurien:
> > this PR should be done, please provide feedback
> 
> @benhepworth would be great if you can consider my comments above ... I'm not yet happy with the PR in the current form. We should also consider to release the API with another initial version instead of declaring it already stable, as the added parameters are more than a minor change and might need some revision in a next version. cc: @eric-murray @jlurien further thoughts?

For the new proposed properties, as they are optional, we should describe what is the meaning or assumption when they are not included as part of a QoS profile, 
- for `l4sQueueType`, would it imply that value `non-l4s-queue` is to be assumed? Or that the usage or not of anyL4S queue is not disclosed.
- for `serviceClass`, there is a `standard` value documented as " Default traffic class", is this a default value to be assumed?

If indeed there are default values to be assumed, we may use the `default` indicator in OAS. But we should be sure that these new parameters make sense for every possible QoS Profile, which I don't know if it's the case. However, if omitting them may imply another meaning, we should make that more explicit in the spec.

RandyLevensalor:
> > > this PR should be done, please provide feedback
> > 
> > 
> > @benhepworth would be great if you can consider my comments above ... I'm not yet happy with the PR in the current form. We should also consider to release the API with another initial version instead of declaring it already stable, as the added parameters are more than a minor change and might need some revision in a next version. cc: @eric-murray @jlurien further thoughts?
> 
> For the new proposed properties, as they are optional, we should describe what is the meaning or assumption when they are not included as part of a QoS profile,
> 
> * for `l4sQueueType`, would it imply that value `non-l4s-queue` is to be assumed? Or that the usage or not of anyL4S queue is not disclosed.
> * for `serviceClass`, there is a `standard` value documented as " Default traffic class", is this a default value to be assumed?
> 
> If indeed there are default values to be assumed, we may use the `default` indicator in OAS. But we should be sure that these new parameters make sense for every possible QoS Profile, which I don't know if it's the case. However, if omitting them may imply another meaning, we should make that more explicit in the spec.

@jlurien Are you sure that we can assume a default value or behavior.  Some operators may not want to expose their queue type or could have an L4S queue, even if the specific profiles doesn't support L4s.

Part of making this optional is that we would not impact operators who are not implementing L4S or operators who are only using DSCP for traffic that isn't managed through QoD, such as voice traffic on 5G networks.  If we define a default behavior, then we are making assumptions and requirements that may not be true about the underlying network.

RandyLevensalor (approved):

jlurien:
> @jlurien Are you sure that we can assume a default value or behavior. Some operators may not want to expose their queue type or could have an L4S queue, even if the specific profiles doesn't support L4s.
> 
> Part of making this optional is that we would not impact operators who are not implementing L4S or operators who are only using DSCP for traffic that isn't managed through QoD, such as voice traffic on 5G networks. If we define a default behavior, then we are making assumptions and requirements that may not be true about the underlying network.

@RandyLevensalor I'm not suggesting to assume any default values, it was just a question. If the answer is that a default value has not to be assumed, which I agree with, we may need to make that more explicit in the description.

Do these new parameters have any incompatibility with or impact on other existing parameters? I'm worried that these new parameters are very technical for an average customer and introduce complexity to the API.

hdamker:
> @RandyLevensalor I'm not suggesting to assume any default values, it was just a question. If the answer is that a default value has not to be assumed, which I agree with, we may need to make that more explicit in the description.
> 
> Do these new parameters have any incompatibility with or impact on other existing parameters? I'm worried that these new parameters are very technical for an average customer and introduce complexity to the API.

We shouldn't forget that this QoSProfiles are defined by API providers to describe the offered profile, they are not meant to be created or defined by API consumers. It is for sure possible to describe profiles which are contradictions in itself, but to avoid that is the responsibility of the provider (as the provider has actually to deliver the profile according to the description).

I share part of the concern that these parameters require deeper network knowledge (and knowledge about the referenced standards), but on the other side is L4S a powerful mechanism for application to adapt to network conditions, which network provider want to expose to application developers. On the ServiceClasses: as they are mainly understood as an indication for which type of traffic the profile is suitable, I'm fine. But they need to be documented in a way that they are not restricting the way of implementation of the profile in different network. But again: the network provider does not need to set the parameter in a profile.

benhepworth:
@hdamker - I updated this PR so that l4sQueueType and serviceClass are experimental in the description.

RandyLevensalor (approved):
LGTM

hdamker (approved):

RandyLevensalor:
@camaraproject/quality-on-demand_codeowners Are we good to merge this?

PR #400, opened 2025-01-21
Use common artifact for device test scenarios

jlurien:
#### What type of PR is this?

* enhancement/feature
* tests

#### What this PR does / why we need it:

Error scenarios for device management are designed in a new artifact in Commonalities, that can be reused. This deprecates previous scenarios testing the same aspects.

README has been updated, removing some technical debt that has been resolved

#### Which issue(s) this PR fixes:

Fixes #398

#### Special notes for reviewers:

To be updated also with outcome of #399 

#### Changelog input

```
Error handling for device management updated to latest guidelines

```

PR #410, opened 2025-01-27
Adapt text on identifying devices from the access token for qos-profiles.yaml

### LINKED ISSUE #399: Adapt logic to identify the device from the access token to QoS Profiles
**Problem description**
We have copied the commonalities text to QoS Profiles, but in that API, device is just an optional input parameter for retrieveQoSProfiles, to filter the available QoS Profiles for certain device, if provided, but device is not mandatory in any case, so current bullet:

> * When the API is invoked using a two-legged access token, the subject will be identified from the optional device object, which therefore MUST be provided.

is wrong, When the operation retrieveQoSProfiles has no input filter, it must return all available QoS Profiles

**Expected action**

An adapted text should replace that section:

```
# Identifying the device from the access token

This operation `retrieveQoSProfiles` allows the API consumer to provide a device as a filter for the operation, as follows:

- When the API is invoked using a two-legged access token, the optional input device object, if present, will be used as filter.

- When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the device to filter will be uniquely identified from the access token.

This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.

## Error handling:
- If the device can be identified from the access token and the optional `device` object is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
```

* We would need to remove the code `MISSING_IDENTIFIER` for 422 responses 

* Codes `SERVICE_NOT_APPLICABLE`  and `UNSUPPORTED_IDENTIFIER`  may be not necessary as well, as an empty response may be returned ([]), meaning no QoS Profiles matching the input criteria, unless we want to make the client aware of the circumstance.

eric-murray:
#### What type of PR is this?
* documentation

Text needs to be adapted as NOT providing a device identifier (either via token or explicit request body parameter) is NOT an error.
- Updates text, indicating that providing both a 3-legged token and explicit device identifier is an error, but providing neither is not
- Remove error `422 MISSING_IDENTIFIER` as this is not a valid error for either endpoint

Fixes #399 

`422 IDENTIFIER_MISMATCH` left in as explicit `device` object may contain mismatched identifiers
`422 SERVICE_NOT_APPLICABLE` left in as some implementations may prefer to indicate that the service is just not available for a specific device rather than saying no profiles are currently available for that device

Test cases not yet updated

#### Changelog input
```
 release-note
 - Updates text, indicating that providing both a 3-legged token and explicit device identifier is an error, but providing neither is not
 - Remove error `422 MISSING_IDENTIFIER` as this is not a valid error for either endpoint
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -26,17 +26,17 @@ info:
 
     # Identifying the device from the access token
 
-    This API requires the API consumer to identify a device as the subject of the API as follows:
-    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
+    The two operations defined by this API (`getQosProfile` and `retrieveQoSProfiles`) both allow the API consumer to use a three-legged access token to specify a target device as a query filter. When provided, only profiles available to the specified target device will be returned. Additionally, when a two-legged acccess token is used, the operation `retrieveQoSProfiles` allows the API consumer to optionally specify the target device in the request body.
 
-    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
+    Hence, for the `retrieveQoSProfiles` operation:
+    - When invoked using a two-legged access token, the optional input device object will be used as filter if present.
+
+    - When invoked using a three-legged access token, this optional identifier in the request body MUST NOT be provided, as the device to filter will be uniquely identified from the access token.
 
     This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
 
     ## Error handling:
-    - If the subject cannot be identified from the access token and the optional `device` object is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
-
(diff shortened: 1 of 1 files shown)


RandyLevensalor (approved):
Good catch.

hdamker (approved):
LGTM. 

@eric-murray do you plan the test case updates within #400?

eric-murray:
> @eric-murray do you plan the test case updates within #400?

Yes, I'll review that PR as well, and approve it when I am finished

hdamker:
> Test cases not yet updated

@eric-murray sorry for pushing, but can we get this done this week? Would be great to get the release PR complete for the RM review.

eric-murray:
Hi @hdamker
The test cases are fixed in #400, which I've approved. So both that and this PR are ready to be merged.

hdamker:
> Hi @hdamker The test cases are fixed in #400, which I've approved. So both that and this PR are ready to be merged.

Ah ... my bad, I misunderstood the comment in #400: "https://github.com/camaraproject/QualityOnDemand/issues/399 will be resolved by https://github.com/camaraproject/QualityOnDemand/pull/410 and not this PR". But I see that @jlurien already did the changes to the test scenarios there.

hdamker (approved):

PR #423, opened 2025-02-21
Update wording in the qod-provisioning API

### LINKED ISSUE #415: Spring25 M3 review comment on terminology used in teh qos-provisioning API
**Problem description**
Raising this issue per comment: https://github.com/camaraproject/QualityOnDemand/pull/407#discussion_r1946508864

A comment on the English used in the description fields in various places:

"provisioning" is not a thing, it is a process. IMHO the item should be a "QoS profile" instead. e.g. "create a new provisioning ... " is a strange sentence.
the text is mixing "QoS profile and "QoD profile" --> should be "QoS profile" consistently
the API names and operationIDs are also a bit strange - e.g. createProvisioning iso requestQosProfile ?
API names:
url = "{apiRoot}/qod-provisioning/v0.2rc1"

the api-names are:
/device-qos (POST)
/device-qos/{provisioningId} (GET, DELETE)
/retrieve-device-qos

**Expected action**

more logically, with the url = {apiRoot}/qod-provisioning/v0.2rc1", the API names could be changed to
/device-qos-profiles
/device-qos-profiles/{qosProfileId}
/retrieve-device-qos-profiles

or shorter to
/qos-profiles
/qos-profiles/{qosProfileId}
/retrieve-qos-profiles

jlurien:
#### What type of PR is this?

* documentation

Being "provisioning" a process and not a "resource", it does not sound well to create or delete a provisioning, but to trigger the process or to revert or revoke the assignments done by a previous provisioning process.

Adresses part of #415 but will not close it

Only some descriptions and operationIds have been adjusted for this Meta, as there is no a clear alternative for the names of the resources used in path, scopes, etc, and this change is critical to be done for this Meta-release without a thoughtful discussion.

As this is still an initial version of the API we will work further on those aspects towards a future meta-release, taking feedback from early integrations.

#### Changelog input

```
- operationIds renamed and some descriptions changed

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -6,7 +6,7 @@ info:
 
     The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId that uniquely identifies this record for later use.
 
-    In addition, this API sets up the configuration in the network such that the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with the provisioned QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
+    Additionally, this API configures the network to apply the requested QoS profile to a specified device whenever the device is connected to the network, until the provisioning is revoked.
 
     # Relevant terms and definitions
 
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -6,7 +6,8 @@ The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for
 
 The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
 
-    This API sets up the configuration in the network so the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with a certain QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
+In addition, this API sets up the configuration in the network such that the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with the provisioned QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
+
 
 
     # Relevant terms and definitions
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -2,13 +2,11 @@ openapi: 3.0.3
 info:
   title: QoD Provisioning API
   description: |
-The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for developers to request the association of a certain QoS profile with a certain device, indefinitely.
+    The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for developers to request the association of a certain QoS profile with a certain device, indefinitely.
 
-The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
+    The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
 
-In addition, this API sets up the configuration in the network such that the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with the provisioned QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
-
-
+    In addition, this API sets up the configuration in the network such that the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with the provisioned QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
 
     # Relevant terms and definitions
 
@@ -16,7 +14,7 @@ In addition, this API sets up the configuration in the network such that the req
     Latency, throughput or priority requirements of the application mapped to relevant QoS profile values. The set of QoS Profiles that a network operator is offering may be retrieved via the `qos-profiles` API (cf. https://github.com/camaraproject/QualityOnDemand/) or will be agreed during the onboarding with the API service provider.
 
     * **Identifier for the device**:
-    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the network operator for the device, at the request time. After the provisioning request is accepted, the device may get different IP addresses, but the provisioning will still apply to the device that was identified during the request process. Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
+    At least one identifier for the device out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the network operator for the device, at the request time. After the provisioning request is accepted, the device may get different IP addresses, but the provisioning will still apply to the device that was identified during the request process. Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
     * **Notification URL and token**:
     API consumers may provide a callback URL (`sink`) on which notifications about all status change events (eg. provisioning termination) can be received from the API provider. This is an optional parameter. The notification will be sent as a CloudEvent compliant message. If `sink` is included, it is RECOMMENDED for the client to provide as well the `sinkCredential` property to protect the notification endpoint.
@@ -232,7 +230,7 @@ paths:
     delete:
       tags:
         - QoD Provisioning
-      summary: Reverts a QoD provisioning
+      summary: Revokes a QoD provisioning
       description: |
         Revokes the association of a QoS profile with the device performed by a previous provisioning.
 
@@ -246,13 +244,13 @@ paths:
         - The access token may be either 2-legged or 3-legged.
         - If a 3-legged access token is used, the subject associated with the QoD provisioning must also be associated with the access token.
         - The QoD provisioning must have been created by the same API consumer given in the access token.
-      operationId: revertProvisioning
+      operationId: revokeProvisioning
       parameters:
         - $ref: "#/components/parameters/provisioningId"
         - $ref: "#/components/parameters/x-correlator"
       responses:
         "204":
-          description: Provisioning reverted
+          description: Provisioning revoked
           headers:
             x-correlator:
               $ref: "#/components/headers/x-correlator"
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -2,7 +2,7 @@ openapi: 3.0.3
 info:
   title: QoD Provisioning API
   description: |
-    The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for developers to request the association of a certain QoS profile with a certain device, indefinitely.
+    The Quality-on-Demand (QoD) Provisioning API provides a programmable interface for developers to request the association of a specific QoS profile with a device, indefinitely.
 
     The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
 
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -234,7 +234,8 @@ paths:
         - QoD Provisioning
       summary: Reverts a QoD provisioning
       description: |
-        Revokes the assignment of a QoS profile to a device performed by a former provisioning.
+        Revokes the association of a QoS profile with the device performed by a previous provisioning.
+
 
         If the notification callback is provided and the provisioning status was `AVAILABLE`, when the process is completed, the API consumer will receive in addition to the response a `PROVISIONING_STATUS_CHANGED` event with
         - `status` as `UNAVAILABLE` and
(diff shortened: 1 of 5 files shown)


RandyLevensalor (changes requested):
Update feature files to match new operationIds.

Use a single term to replace delete.  I suggest using `revoke` and not using `revert`.

eric-murray:
Changing the `operationId` will require an update to associated the feature file

code/API_definitions/qod-provisioning.yaml
```diff
@@ -396,7 +396,7 @@ components:
     ProvisioningInfo:
       description: |
         Provisioning related information returned in responses.
-        Optional device object only to be returned if provided in createProvisioning. If more than one type of device identifier was provided, only one identifier will be returned (at implementation choice and with the original value provided in createProvisioning).
+        Optional device object only to be returned if provided in triggerProvisioning. If more than one type of device identifier was provided, only one identifier will be returned (at implementation choice and with the original value provided in triggerProvisioning).
         Please note that IP addresses of devices can change and get reused, so the original values may no longer identify the same device. They identified the device at the time of QoD provisioning creation.
```
tanjadegroot, line 400:
```suggestion
        Please note that IP addresses of devices can change and get reused, so the original values may no longer identify the same device. They identified the device at the time the QoD provisioning was triggered.

```

tanjadegroot (approved):
Some further suggestions and it is OK for me. thanks !

hdamker:
I'm actually not happy with the PR. 

The name of the resources to create, read and delete should be a noun which is describing the kind of resource which is created. I agree that "Provisioning" isn't that noun, as it is not describing the outcome but the activity to get there ("Provision" was actually better in this sense, as it could be understood as the outcome of the activity of provisioning). Options are for example "QosProfileAssociation" or shorter “QosAssignment”, “ProfileApplication”, or even "QosProvision".

The attempt to circumvent this needed change with new operation ids like "triggerProvisioning" goes into the wrong direction, as these changes deviate from RESTful and need to be reverted when later resource name is replaced with a proper noun. Also "revokeProvisioning" is a term which doesn't make sense to me (especially if the event is still "DELETE_REQUESTED").

I would rather propose to keep the API structure for now as is, but add a clear disclaimer that the resource name "Provisioning" is for now only a placeholder, should be read in the sense as described above (as the outcome of the provisioning process), and will be replaced in the next version with a better term (as a breaking change).

Unfortunately this Initial API is within the QualityOnDemand repository, otherwise my proposal would be to skip for qos-provisioning the Spring25 meta-release, and prepare the needed changes with enough time for Fall25. Don't know if carving the API out of the repository short-term is still an option? For me is the issue here a good example that APIs in their first initial versions shouldn't be in an Incubating repository but having their own Sandbox, at least in the initial phase.

tanjadegroot:
> Note: I replaced my complete previous comment. The reason is that I hadn't realized that the API has actually a resource noun (`device-qos`) which makes sense if understood as a short for "device-qos-assignment" or "device-qos-association" or even "device-qos-provision" (the result of a provisioning). The challenge is that this short form has no plural form.

"device-qos-association" would be the right resource name for this API IMHO. It took me also a while to understand that that is the actual resource being managed through this API. 

> 
> Beyond that the problems to address are more the Schemata names (which can be replaced without impact on API clients), the parameter `provisioningId` (which can't be changed without a breaking change), and beyond that mostly the descriptions and operationIds which are not referring to the resource which get created, read and deleted. I agree that this shouldn't be done short-term.
> 
> The operationIds need to be changed again later, as they should follow the chosen resource name.
> 
> In summary I'm fine with the PR, but would propose to add some disclaimer, that it is an initial version and a later version will rename certain operationIds, schemata names and also property names to make it more consistent regarding the actual resources.

I would also plan to renaming of the resource to "device-qos-association" as well, as anyway you are doing breaking changes. it will much clarify the usage of the API.

> 
> BTW: If the API name should be qos-provisioning or qod-provisioning is also worth a discussion. A QoS profile will be provisioned, but not a "QoD" (which is the name of another API which is about has `sessions` as resource) ... the title of this PR is using "qos-provisioning API" ;-)

I agree the usage of "qod" is confusing and should be "qos"

hdamker:
> Changing the `operationId` will require an update to associated the feature file

@jlurien could you address that within the next two days, so that we get ready for Feb 28th?

code/API_definitions/qod-provisioning.yaml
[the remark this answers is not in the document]
jlurien: - Profile -> profile
- UE removed

jlurien:
> Changing the `operationId` will require an update to associated the feature file

Updated test plan aligned with the new operationIds and some typos fixed as well

jlurien:
> Some further suggestions and it is OK for me. thanks !

I like your suggestions, I think they bring more clarity. Thank you!

jlurien:
> Note: I replaced my complete previous comment. The reason is that I hadn't realized that the API has actually a resource noun (`device-qos`) which makes sense if understood as a short for "device-qos-assignment" or "device-qos-association" or even "device-qos-provision" (the result of a provisioning). The challenge is that this short form has no plural form.
> 
> Beyond that the problems to address are more the Schemata names (which can be replaced without impact on API clients), the parameter `provisioningId` (which can't be changed without a breaking change), and beyond that mostly the descriptions and operationIds which are not referring to the resource which get created, read and deleted. I agree that this shouldn't be done short-term.
> 
> The operationIds need to be changed again later, as they should follow the chosen resource name.
> 
> In summary I'm fine with the PR, but would propose to add some disclaimer, that it is an initial version and a later version will rename certain operationIds, schemata names and also property names to make it more consistent regarding the actual resources.
> 
> BTW: If the API name should be qos-provisioning or qod-provisioning is also worth a discussion. A QoS profile will be provisioned, but not a "QoD" (which is the name of another API which is about has `sessions` as resource) ... the title of this PR is using "qos-provisioning API" ;-)

I agree that we should rethink about resources and names towards the next meta. Regarding the disclaimer, I assume that that is implicit to an API in initial version. I prefer to think first in what to change and eventually inform in a later meta that we have changed something.

jlurien:
> > Changing the `operationId` will require an update to associated the feature file
> 
> @jlurien could you address that within the next two days, so that we get ready for Feb 28th?

Please review if all the comments are addressed. I only did not apply the suggestion for a disclaimer, as I'm not sure where and how to apply that, as no other initial API version has ever had such a disclaimer and we have changed API names, resources and operationIds in other cases.

RandyLevensalor:
> Please review if all the comments are addressed. I only did not apply the suggestion for a disclaimer, as I'm not sure where and how to apply that, as no other initial API version has ever had such a disclaimer and we have changed API names, resources and operationIds in other cases.

@jlurien We applied a disclaimer about some of the l4s attributes in the QoS profile.  Here's a link to one of them as an example. https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_definitions/qos-profiles.yaml#L337

hdamker:
> Please review if all the comments are addressed. I only did not apply the suggestion for a disclaimer, as I'm not sure where and how to apply that, as no other initial API version has ever had such a disclaimer and we have changed API names, resources and operationIds in other cases.

@jlurien First of all: I don't insist on the disclaimer, and I see that the suggestions from @tanjadegroot are applied, so I will approve the PR, just need to go over the result and see if it looks consistent to me. 

I get your argument about the initial version - I just think that the situation compared to other APIs is different as we have already at the time of the meta-release a plan for a significant redesign, which I'm not aware of from other examples. One way out of this could be to have soon an alpha version with the changes, so that implementors can decide if they spent the effort for 0.2.0 or are waiting for the next iteration of the API. To allow such different pace and degree of change we should also think about moving qod-provisioning into an own repository.

code/API_definitions/qod-provisioning.yaml
```diff
@@ -2,18 +2,19 @@ openapi: 3.0.3
 info:
   title: QoD Provisioning API
   description: |
-    The Quality-On-Demand (QoD) Provisioning API offers a programmable interface for developers to request the assignment of a certain QoS Profile to a certain device, indefinitely.
+    The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for developers to request the association of a certain QoS profile with a certain device, indefinitely.
```
hdamker, line 5:
```suggestion
    The Quality-on-Demand (QoD) Provisioning API provides a programmable interface for developers to request the association of a specific QoS profile with a device, indefinitely.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -2,18 +2,19 @@ openapi: 3.0.3
 info:
   title: QoD Provisioning API
   description: |
-    The Quality-On-Demand (QoD) Provisioning API offers a programmable interface for developers to request the assignment of a certain QoS Profile to a certain device, indefinitely.
+    The Quality-on-Demand (QoD) Provisioning API offers a programmable interface for developers to request the association of a certain QoS profile with a certain device, indefinitely.
 
-    This API sets up the configuration in the network so the requested QoS profile is applied to a specified device, at any time while the provisioning is available. The device traffic will be treated with a certain QoS profile by the network whenever the device is connected to the network, until the provisioning is deleted.
+    The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
```
hdamker, line 7:
```suggestion
    The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId that uniquely identifies this record for later use.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -2,18 +2,19 @@ openapi: 3.0.3
...
 
-    This API sets up the configuration in the network so the requested QoS profile is applied to a specified device, at any time while the provisioning is available. The device traffic will be treated with a certain QoS profile by the network whenever the device is connected to the network, until the provisioning is deleted.
+    The association resulting from the QoD provisioning request is represented by a QoD provisioning record (or QoD provisioning for short) that includes information about the date of provisioning, the QoS profile, the provisioning status, etc., as well as a provisioningId uniquely identifying this record for later usage.
 
+    In addition, this API sets up the configuration in the network such that the requested QoS profile is applied to a specified device, at any time while the association is available. The device traffic will be treated with the provisioned QoS profile by the network whenever the device is connected to the network, until the provisioning is revoked.
```
hdamker, line 9: Copilot suggests here a significant simplification which I like:

```suggestion
    Additionally, this API configures the network to apply the requested QoS profile to a specified device whenever the device is connected to the network, until the provisioning is revoked.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -24,7 +25,7 @@ info:
     - An operation to setup a new QoD provisioning for a given device.
```
hdamker, line 25:
```suggestion
    - An operation to set up a new QoD provisioning for a given device.
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -82,9 +83,10 @@ paths:
     post:
       tags:
         - QoD Provisioning
-      summary: Sets a new provisioning of QoS for a device
+      summary: Triggers a new provisioning of QoS for a device
       description: |
-        Triggers a new provisioning in the operator to assign certain QoS Profile to certain device.
+        Triggers the provisioning in the operator's environment of the association of a given QoS profile with a given device.
```
hdamker, line 88: Copilot suggests to revert part of the change ... ignore if you want to introduce already the term "association":

```suggestion
        Triggers the provisioning in the operator's environment to associate a given QoS profile with a given device.
```

hdamker (approved):
The PR looks LGTM, here just a few selection from the suggestions I got from Copilot, which I found valuable for readability.

jlurien:
> > Please review if all the comments are addressed. I only did not apply the suggestion for a disclaimer, as I'm not sure where and how to apply that, as no other initial API version has ever had such a disclaimer and we have changed API names, resources and operationIds in other cases.
> 
> @jlurien We applied a disclaimer about some of the l4s attributes in the QoS profile. Here's a link to one of them as an example. https://github.com/camaraproject/QualityOnDemand/blob/main/code/API_definitions/qos-profiles.yaml#L337

This is a note on the description of a specific new property introduced in that version, not a general disclaimer advising that paths, operationIds and even the API name can change in a future version.

jlurien:
> > Please review if all the comments are addressed. I only did not apply the suggestion for a disclaimer, as I'm not sure where and how to apply that, as no other initial API version has ever had such a disclaimer and we have changed API names, resources and operationIds in other cases.
> 
> @jlurien First of all: I don't insist on the disclaimer, and I see that the suggestions from @tanjadegroot are applied, so I will approve the PR, just need to go over the result and see if it looks consistent to me.
> 
> I get your argument about the initial version - I just think that the situation compared to other APIs is different as we have already at the time of the meta-release a plan for a significant redesign, which I'm not aware of from other examples. One way out of this could be to have soon an alpha version with the changes, so that implementors can decide if they spent the effort for 0.2.0 or are waiting for the next iteration of the API. To allow such different pace and degree of change we should also think about moving qod-provisioning into an own repository.

Understood. With the new explanations I think that the concepts are much more clear now. The changes being discussed are rather aesthetic, and do not affect that much the logic of the implementation behind, but as soon as we have a decision about better naming we can create a 0.3.0-alpha.1. 

Now that we have a different repo for qos-booking, it could make sense to have a separate repo for the provisioning one, as well. We probably had decided to do so initially, if we had the process defined at the time.

code/API_definitions/qod-provisioning.yaml
jlurien: I prefer the new one, more concise.

jlurien:
Thanks for the suggestions @hdamker. About the posible disclaimer. If you think it bring value, we could add it as a note to the readme or release notes while creating the public release PR

jlurien:
@hdamker let me know if anything else is needed here for the public release PR. I think we still need @RandyLevensalor's approval as he requested changes

RandyLevensalor (approved):
That looks much better @jlurien

hdamker (approved):

eric-murray (approved):

hdamker:
So we are done here :-)

PR #434, opened 2025-04-24
Adding new query atttributes for retrieve QoS profile

### LINKED ISSUE #147: Extend the query capabilities for Qos Profiles
**Problem description**
The initial implementation of the   Get `/qos-profiles:` endpoint is fairly simple.
Developers will want a richer solution to find the appropriate profile.

**Possible evolution**
There are several methods to achieve this.  One or more can be implemented.

1. Extend the current query parameters to allow for comparison operators =,<,>,>=, or <= that would follow the commonalities guidelines.  For instance  <attribute>.gt=<value> would return all profiles where attribute is larger than the given value.
2. Serialize complex types, such as Rate to be a comma delimited string.  For example "10,Mbps" could be used to signify a rate of 10Mbps. For Duration 10,ms would denote 10 milliseconds.
3. Not all profiles will be available for a given device or user.  For instance users may not be entitled to the servers based on their current plan or the access network is not capable of providing certain levels of service.  For this we can add device and flow identifiers to the query.  These would follow the same format used to identify devices in the`CreateSession:` schema.
 
**Alternative solution**
A json string could be passed instead of a comma delimited complex query parameter. 

**Additional context**
This functionality was remove from the 0.9.0 release in order to provide additional time to discuss and design this endpoint.

RandyLevensalor:
#### What type of PR is this?

* enhancement/feature

Adds new query capabilites

Fixes #147

#### Changelog input

```
 release-note
Enhanced filtering capabilities in the /retrieve-qos-profiles endpoint:
Added support for filtering by all QoS profile attributes including rates, durations, priority, latency, and service classes
Clarified comparison logic for different types of attributes:
Rate attributes: returned profiles have values greater than or equal to requested
Duration attributes:
minDuration: profiles have values less than or equal to requested
maxDuration: profiles have values greater than or equal to requested
Priority: profiles have values less than or equal to requested (lower = higher priority)
Latency attributes (jitter, packetDelayBudget): profiles have values less than or equal to requested
packetErrorLossRate: profiles have values less than or equal to requested
l4sQueueType and serviceClass: profiles match exactly

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -451,7 +451,15 @@ components:
 
     QosProfileDeviceRequest:
       description: |
-        Request object for QoS Profiles for a given device
+        Request object for QoS Profiles for a given device.
+        When attributes are specified, only profiles that meet or exceed the requested capabilities will be returned:
+        - Rate attributes (targetMinUpstreamRate, maxUpstreamRate, etc.): returned profiles will have rates greater than or equal to the requested values
+        - Duration attributes:
+          - minDuration: returned profiles will have minDuration less than or equal to the requested value
+          - maxDuration: returned profiles will have maxDuration greater than or equal to the requested value
+        - Priority: returned profiles will have priority less than or equal to the requested value (lower values = higher priority)
+        - Jitter & packetDelayBudget: returned profiles will have values less than or equal to the requested values
+        - packetErrorLossRate: returned profiles will have values less than or equal to the requested value
       type: object
       properties:
         device:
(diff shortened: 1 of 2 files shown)


JoachimDahlgren:
One one side I appreciate that you are able to filter on any of the attributes in the profile. On the other side from the working group meeting today I got the impression that very few of these attributes are set on QoS Profiles today. If that is the case adding a filter will only result in getting nothing back in the response. 

Do we have any indication of information always set in the profile and if so will it make sense to start with enabling filtering only on those?

Also what makes sense from an API invoker perspective. What would be most important to filter on?

RandyLevensalor:
> One one side I appreciate that you are able to filter on any of the attributes in the profile. On the other side from the working group meeting today I got the impression that very few of these attributes are set on QoS Profiles today. If that is the case adding a filter will only result in getting nothing back in the response.
> 
> Do we have any indication of information always set in the profile and if so will it make sense to start with enabling filtering only on those?
> 
> Also what makes sense from an API invoker perspective. What would be most important to filter on?

@JoachimDahlgren thanks for the feedback.  I don't think that the API should limit what can be filtered.  These decisions should be left to the API provider.

I think we provide a mechanism and documentations to clearly state how to handle when attributes are not set.

One option would be to return an error 422 query attribute not supported.  We could also add a new get call to get a list of supported query attributes to better automate the process.

Operators could also provide a default value for attributes that are not customizable through the API.

eric-murray:
> I tend to lean towards a simpler API call, which would have this flag as optional and to default to only returning matching profiles. I'm open to feedback on this point.

I agree - having a default value implicitly makes it an optional parameter (for the API consumer), and I agree the default should be to narrow the range of responses to profiles for which values are specified for the filter parameters

eric-murray:
An additional comment. For documentation, I'd prefer if the applicability of a given filter parameter (i.e. is it and exact match, or greater or equal to, or less than or equal to) be specified in a description property for each individual parameter, rather than being in the overall description for the request body. This makes it easier for the API consumer to understand how the filter will be applied when they are looking at the parameter itself.

This can be done using the `allOf` trick we use elsewhere. e.g. something like:
```
    QosProfileDeviceRequest:
      description: <short summary>
      type: object
      properties:
        targetMinUpstreamRate:
          description: returned profiles will have a target minimum upstream data rate greater than or equal to the requested value
          allOf:
            -  $ref: "#/components/schemas/Rate"
```

hdamker (commented):
I have to admit that I'm not in favor to add this functionality at all at the current time -- it will make the implementation of the API more complex, hence slowing the adoption across operators, and has from my perspective currently limited value for API consumer. 

We have no experience yet how profiles will actually look like and how many will be there. As we have omitted pagination, as the general assumption is that it will be a small number.

If we add the functionality there is at least one change needed.

hdamker:
As decided within Sub Project call on June 13th: Changing the PR to draft and backlog

PR #435, opened 2025-04-25
Add country availability field to QoS profile

### LINKED ISSUE #431: Add "countryAvailability" property to QoS Profiles
**Problem description**
Currently QoS Profiles have no indication as to which countries they are available in. Because network operators may have networks in multiple countries, it may be that a given QoS profile is available in one country but not another.

Whilst the country of interest might be inferred from the device filter (or 3-legged access token) if used, or from the country in which the API consumer operates, it is still possible that the API consumer might want to know in which countries a given QoS profile is available for any device.

Currently, an API provider would have to create separate QoS profiles for each country in which they operate, even if other parameters of the profile are the same. Whilst the country name could be incorporated into the QoS profile name, this would not allow for reliable filtering.

**Possible evolution**
Add a `countryAvailability` property to the QoS profile. This would be an array of 2-letter ISO country codes.

e.g. availability in UK, Germany, France, Italy and Spain could be indicated as follows:
```
...
"countryAvailablility": [ "GB", "DE", "FR", "IT", "ES" ],
...
```

The `countryAvailability` property would be optional. If omitted, the profile is available in all countries where the network operator operates.

**Alternative solution**
None proposed

**Additional context**
None

eric-murray:
#### What type of PR is this?
* enhancement/feature

Adds a `countryAvailability` field to the QoS profile to allow an API provider to indicate in which countries the profile is available

Fixes #431 

None

#### Changelog input

```
 release-note
 - Adds a countryAvailability field to the QoS profile to allow an API provider to indicate in which countries the profile is available
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -204,6 +204,18 @@ components:
             A description of the QoS profile.
           type: string
           example: "QoS profile for video streaming"
+        countryAvailablility:
+          description: A list of countries in which the API provider makes the profile available.
+                       - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to the API provider's network.
+                       - When there is no device filter, availability in a given country means that the profile will be available to some devices connected to the API provider's network, but not necessarily to all devices.
+                       - If this field is not present in the QoS profile, no assumption can be made about the availability of the profile in any given country
+          type: array
+            items:
+              type: string
+              description: The two letter ISO 3166-2 country code for the country in which the QoS profile is available
+              pattern: '^[A-Z]{2}$'
+              example: "GB"
+            example: "[ "GB", "DE", "FR", "IT", "ES" ]"
         status:
           $ref: "#/components/schemas/QosProfileStatusEnum"
         targetMinUpstreamRate:
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -207,10 +207,11 @@ components:
         status:
           $ref: "#/components/schemas/QosProfileStatusEnum"
         countryAvailablility:
-          description: A list of countries in which the API provider makes the profile available.
-                       - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to the API provider's network.
-                       - When there is no device filter, availability in a given country means that the profile will be available to some devices connected to the API provider's network, but not necessarily to all devices.
-                       - If this field is not present in the QoS profile, no assumption can be made about the availability of the profile in any given country
+          description: |
+            A list of countries in which the API provider makes the profile available.
+            - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to the API provider's network.
+            - When there is no device filter, availability in a given country means that the profile will be available to some devices connected to the API provider's network, but not necessarily to all devices.
+            - If this field is not present in the QoS profile, no assumption can be made about the availability of the profile in any given country
           type: array
           items:
             type: string
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -204,6 +204,8 @@ components:
             A description of the QoS profile.
           type: string
           example: "QoS profile for video streaming"
+        status:
+          $ref: "#/components/schemas/QosProfileStatusEnum"
         countryAvailablility:
           description: A list of countries in which the API provider makes the profile available.
                        - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to the API provider's network.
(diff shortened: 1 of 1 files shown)


hdamker:
@eric-murray just now checked [device-roaming-status.yaml](https://github.com/camaraproject/DeviceStatus/blob/main/code/API_definitions/device-roaming-status.yaml) they are using:

```
CountryCode:
      description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
      type: integer

    CountryName:
      description: The ISO 3166 ALPHA-2 country-codes of mapped to mobile country code(MCC). If there is mapping of one MCC to multiple countries, then we have list of countries. If there is no mapping of MCC to any country, then an empty array [] shall be returned..
      type: array
      items:
        type: string
```

So maybe we should use `countryName` instead of `country`?

And they are providing both for a device :
```
        countryCode:
          $ref: "#/components/schemas/CountryCode"
        countryName:
          $ref: "#/components/schemas/CountryName"
```

Looking on the definition of `CountryName`, I'm not sure that every if is always possible to have a 1:m relation between country and networks. Could it happen that you need a m:n one? So a list of countryNames on the one side and one or more networks on the other side (in case one or more covered networks have an MNC which is spanning multiple countries?).

code/API_definitions/qos-profiles.yaml
```diff
@@ -206,6 +206,14 @@ components:
           example: "QoS profile for video streaming"
         status:
           $ref: "#/components/schemas/QosProfileStatusEnum"
+        countryAvailability:
+          description: |
+            A list of countries, and optionally networks, for which the API provider makes the profile available.
+            - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to any of the listed networks, or the API provider's network if none are listed
```
jlurien, line 212: We may need to clarify that the service will work in roaming mode for this device in the listed countries/networks outside of the home network.

code/API_definitions/qos-profiles.yaml
```diff
@@ -206,6 +206,14 @@ components:
...
+            - When there is no device filter, availability for a given country means that the profile will be available to some devices connected to the networks listed for that country (or the API provider's network if none are listed), but not necessarily available to all devices, and not necessarily available to the same device for all listed countries and networks
```
jlurien, line 213: In this case, it is not clear if the service is available in roaming mode in some cases, or only available to home devices in the listed countries/networks. In combination with the point above, it may be a bit misleading.

eric-murray: The proposal is that, if no specific device is identified, then the API cannot be used to determine if it is possible for a device to use the QoS profile on all listed networks (through roaming). Maybe it can, maybe it can't, maybe it works for some devices but not others. But the API consumer will have to find out some other way, ideally by querying the API for a specific device.

Maybe you could propose some alternative text to make that clearer, always remembering that not all networks are mobile networks!

code/API_definitions/qos-profiles.yaml
```diff
@@ -206,6 +206,14 @@ components:
           example: "QoS profile for video streaming"
         status:
           $ref: "#/components/schemas/QosProfileStatusEnum"
+        countryAvailability:
+          description: |
+            A list of countries, and optionally networks, for which the API provider makes the profile available.
+            - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to any of the listed networks (assuming roaming is permitted for networks that are not the device's home network), or the API provider's own network if no networks are listed
+            - When there is no device filter, availability for a given country means that the profile will be available to some devices connected to the networks listed for that country (or the API provider's network if none are listed), but not necessarily available to all devices, and not necessarily available to the same device for all listed countries and networks
```
jlurien, line 213:
```suggestion
            - When there is no device filter, availability for a given country means that the QoS profile may be available to some devices connected to the listed networks in that country (or to the API provider’s default network if none are listed). However, this does not imply that the profile is available to all devices, or that a particular device will be able to use the profile across all listed countries or networks. In particular, the response does not provide enough information to determine availability for roaming for specific devices; consumers should query the API with a specific device identifier to get definitive availability.
```

code/API_definitions/qos-profiles.yaml
```diff
@@ -206,6 +206,14 @@ components:
           example: "QoS profile for video streaming"
         status:
           $ref: "#/components/schemas/QosProfileStatusEnum"
+        countryAvailability:
+          description: |
+            A list of countries, and optionally networks, for which the API provider makes the profile available.
+            - When profiles are filtered for a specific device (i.e. using a 3-legged access token or an explicit device filter parameter), the profile can be used by that device in any of these countries when connected to any of the listed networks (assuming roaming is permitted for networks that are not the device's home network), or the API provider's own network if no networks are listed
```
hdamker, line 212: I try to remember if we have discussed that already, but shouldn't the API request return only profiles which can be actually used by the device within the network where it is currently connected to? Taking for example the available network technologies into consideration)?

hdamker: In other words: if asked for a specific device it might be better not to provide the field at all.

hdamker (commented):
I still think that the field "countryAvailablity" is (mainly/only?) useful if aggregators are asked for available profiles without a specific device (e.g. the request can't be routed to a specific network and the aggregator can only provide an "aggregated" result or their own service portfolio). An operator group aggregating multiple networks falls into the same category.

hdamker (approved):

hdamker:
@jlurien @RandyLevensalor could you make your review before the QoD meeting tomorrow?

RandyLevensalor (approved):

hdamker:
Great, we are done here. @eric-murray will you merge the PR?

PR #447, opened 2025-06-13
Refine terminology in qos-provisioning API

jlurien:
#### What type of PR is this?

* cleanup
* documentation

#### What this PR does / why we need it:

The API specification has undergone several significant improvements to address terminology issues and enhance clarity:

* Title changed from "QoD Provisioning API" to "QoS Provisioning", updating all related tags, scopes, urls, eventTypes etc
* Better clarify that "provisioning" is a process, not a thing
* Paths changed: 
  -  `/device-qos` to `/qos-assignments`
  - `/retrieve-device-qos` to `/retrieve-qos-assignment`
* Updated operation IDs to be more descriptive (e.g., triggerQosProvisioning)
* Improved endpoint summaries and descriptions

#### Which issue(s) this PR fixes:

Fixes #415 

#### Special notes for reviewers:

- **Filename has not been renamed so changes are easier to detect and review, but if we agree on the new name we should later rename it**

#### Changelog input

```
- API renamed to QoS Provisioning
- URLs and paths changed
- Descriptions enhanced 
```

PR #453, opened 2025-06-17
Update OAS definitions to remove IDENTIFIER_MISMATCH error code

### LINKED ISSUE #450: Remove IDENTIFIER_MISMATCH error and associated test cases
**Problem description**
Commonalities has removed the `IDENTIFIER_MISMATCH` error for Fall25

**Expected behaviour**
- `422 IDENTIFIER_MISMATCH` must not be a possible error response.
- The `device` property in responses should be of type `DeviceResponse` to limit the number of identifiers that can be included in the response to 1.
- Update test cases and documentation

**Alternative solution**
None

**Additional context**
This is to prevent the API consumer abuse when trying to correlate different identifiers for a given end user

eric-murray:
#### What type of PR is this?
* correction

Commonalities has removed the `IDENTIFIER_MISMATCH` error for Fall25. Instead, a device identifier should not normally be included in responses. It should only be included to disambiguate the identity of the device when:
- the endpoint is being accessed using a 2-legged access token; and
- the endpoint allows for a device identifier to be provided in the request; and
- the API consumer has included multiple device identifiers

The response then includes the device identifier that was used by the implementation

This PR:
- Updates documentation in each OAS
- Adds relevant example both for when device disambiguation is required, and when it is not required
- Uses `DeviceResponse` object to limit device identifiers in responses to exactly one
- Removes the IDENTIFIER_MISMATCH error code option from 422 responses and associated test cases

Fixes #450 

- Response examples could be improved in general. The "default" response examples created by the OAS viewers are generally not typical responses.
- For `qos-profiles`, I included an optional `device` parameter in all profiles returned by the `/retrieve-qos-profiles` endpoint. This is clumsy, but has the advantage of being a non-breaking change (for clients). A better solution would be to have a separate `device` parameter common to all profiles returned, but this would be a breaking change.
- I think all the `GET` and `DELETE` endpoints should never return a device identifier in the response, but haven't thought through all possible scenarios, so will raise this as a separate issue.

#### Changelog input

```
 release-note
 - Update documentation in each OAS
 - Add relevant example both for when device disambiguation is required, and when it is not required
 - Use DeviceResponse object to limit device identifiers in responses to exactly one
 - Remove the IDENTIFIER_MISMATCH error code option from 422 responses and associated test cases
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -1107,7 +1107,7 @@ components:
         qosProfile: QOS_L
         sink: https://application-server.com/callback
         provisioningId: 3fa85f64-5717-4562-b3fc-2c963f66afa6
-        qosStatus: REQUESTED
+        status: REQUESTED
 
     PROVISIONING_CREATION_EXAMPLE_WITH_DEVICE_DISAMBIGUATION:
       summary: Successful provisioning creation with device disambiguation
@@ -1118,7 +1118,7 @@ components:
         qosProfile: QOS_L
         sink: https://application-server.com/callback
         provisioningId: 3fa85f64-5717-4562-b3fc-2c963f66afa6
-        qosStatus: REQUESTED
+        status: REQUESTED
 
     PROVISIONING_AVAILABLE:
       summary: QoD provisioning status is available with device disambiguation
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -166,10 +166,10 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProvisioningInfo"
               examples:
-                Successful Provisioning Creation:
-                  $ref: "#/components/examples/PROVISIONING_CREATION_EXAMPLE"
-                Successful Provisioning Creation with Device in the Response:
-                  $ref: "#/components/examples/PROVISIONING_CREATION_EXAMPLE_WITH_DEVICE_RESPONSE"
+                Successful QoS Profile Assignment Creation:
+                  $ref: "#/components/examples/ASSIGNMENT_CREATION_EXAMPLE"
+                Successful QoS Profile Assignment Creation with Device in the response:
+                  $ref: "#/components/examples/ASSIGNMENT_CREATION_EXAMPLE_WITH_DEVICE_RESPONSE"
         "400":
           $ref: "#/components/responses/TriggerProvisioning400"
         "401":
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -214,8 +214,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProvisioningInfo"
               examples:
-                PROVISIONING_AVAILABLE:
-                  $ref: "#/components/examples/PROVISIONING_AVAILABLE"
                 PROVISIONING_UNAVAILABLE:
                   $ref: "#/components/examples/PROVISIONING_UNAVAILABLE"
                 PROVISIONING_AVAILABLE_WITHOUT_DEVICE:
@@ -324,6 +322,8 @@ paths:
               examples:
                 PROVISIONING_AVAILABLE:
                   $ref: "#/components/examples/PROVISIONING_AVAILABLE"
+                PROVISIONING_UNAVAILABLE:
+                  $ref: "#/components/examples/PROVISIONING_UNAVAILABLE"
                 PROVISIONING_AVAILABLE_WITHOUT_DEVICE:
                   $ref: "#/components/examples/PROVISIONING_AVAILABLE_WITHOUT_DEVICE"
         "400":
@@ -1120,7 +1120,7 @@ components:
 
     PROVISIONING_AVAILABLE:
       summary: QoD provisioning status is available
-      description: The provisioning has become available
+      description: The provisioning has become available and device disambiguation is required
       value:
         device:
           phoneNumber: "+123456789"
@@ -1152,7 +1152,7 @@ components:
         statusInfo: NETWORK_TERMINATED
 
     PROVISIONING_AVAILABLE_WITHOUT_DEVICE:
-      summary: QoD provisioning status is available but no device information is provided
+      summary: QoD provisioning status is available
       description: Device is optional in responses and must not be provided if it was not provided in the request
       value:
         qosProfile: QOS_M
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -14,7 +14,7 @@ info:
     Latency, throughput or priority requirements of the application mapped to relevant QoS profile values. The set of QoS Profiles that a network operator is offering may be retrieved via the `qos-profiles` API (cf. https://github.com/camaraproject/QualityOnDemand/) or will be agreed during the onboarding with the API service provider.
 
     * **Identifier for the device**:
-    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone Number, or Network Access Identifier assigned by the mobile network operator for the device. Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the session details. If only a single device identifier is provided, the implementation may or may not include the device identifier in the session details.
+    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone Number, or Network Access Identifier assigned by the mobile network operator for the device. Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the provisioning details. If only a single device identifier is provided, the implementation may or may not include the device identifier in the provisioning details.
 
       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
@@ -1101,7 +1101,7 @@ components:
 
   examples:
     PROVISIONING_CREATION_EXAMPLE:
-      summary: Successful provisioning creation
+      summary: Successful provisioning creation without device in the provisioning details
       description: Example response after a new QoS provisioning has been created using a 3-legged access token, or possibly a single device identifier
       value:
         qosProfile: QOS_L
@@ -1110,7 +1110,7 @@ components:
         status: REQUESTED
 
     PROVISIONING_CREATION_EXAMPLE_WITH_DEVICE_RESPONSE:
-      summary: Successful provisioning creation with device disambiguation
+      summary: Successful provisioning creation with device in the provisioning details
       description: Example response after a new QoS provisioning has been created using a 2-legged access token with multiple device identifiers, or possibly only a single device identifier
       value:
         device:
@@ -1121,8 +1121,8 @@ components:
         status: REQUESTED
 
     PROVISIONING_AVAILABLE:
-      summary: QoD provisioning status is available with device disambiguation
-      description: The provisioning has become available, but multiple device identifiers were provided and hence device disambiguation is required
+      summary: QoD provisioning status is available with device in the provisioning details
+      description: The provisioning has become available, and the device to which the provisioning applies is included in the provisioning details
       value:
         device:
           phoneNumber: "+123456789"
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -14,7 +14,7 @@ info:
     Latency, throughput or priority requirements of the application mapped to relevant QoS profile values. The set of QoS Profiles that a network operator is offering may be retrieved via the `qos-profiles` API (cf. https://github.com/camaraproject/QualityOnDemand/) or will be agreed during the onboarding with the API service provider.
 
     * **Identifier for the device**:
-    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device. Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the session creation response.
+    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone Number, or Network Access Identifier assigned by the mobile network operator for the device. Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the session details. If only a single device identifier is provided, the implementation may or may not include the device identifier in the session details.
 
       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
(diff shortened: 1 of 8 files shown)


code/API_definitions/quality-on-demand.yaml
```diff
@@ -546,6 +560,10 @@ components:
     CreateSession:
       description: Attributes required to create a session
       allOf:
+        - type: object
+          properties:
+            device:
```
jlurien, line 565: idem

code/API_definitions/quality-on-demand.yaml
```diff
@@ -504,6 +511,13 @@ components:
         Note that the device object is defined as optional and will only to be returned if provided in createSession. If more than one type of device identifier was provided, only one identifier will be returned (at implementation choice and with the original value provided in createSession).
         Please note that IP addresses of devices can change and get reused, so the original values may no longer identify the same device. They identified the device at the time of session creation.
       allOf:
+        - type: object
+          properties:
+            device:
```
jlurien, line 516: any reason to separate this property in another item of the `allOf`?

jlurien (commented):
Looks good, with some comments

code/API_definitions/quality-on-demand.yaml
eric-murray: Hi @jlurien 

This was to try and preserve the property order for the documentation and examples automatically generated by the OAS editors / viewers. I appreciate that the order in which properties are listed in a response does not actually matter, but moving properties around in the documentation / examples may give the impression that the changes are more substantial than they actually are.

If it is decided this is not an issue, I can include the `device` property in the existing object within the `allOf`.

code/API_definitions/quality-on-demand.yaml
eric-murray: idem

code/API_definitions/quality-on-demand.yaml
eric-murray: Well, of course, that would work, in the same way that we could remove the `BaseSessionInfo` and just copy its contents directly into `SessionInfo` and the other places it is used. Having a dedicated `DeviceResponse` schema means that any future changes only need one update per YAML, rather than having to search out everywhere that the `Device` schema has been extended as proposed above.

So we are trying to strike a balance between maintainability and implementability. Are there no tools that can search out `allOf`s and replace them directly with the inherited properties for the benefit of code generators? Why don't code generators do that as a first step?

I think if we need additional rules on OAS construction beyond the OAS specification itself (and the "OK" from the OAS viewers) for the benefit of code generators, these rules really need to be properly documented.

maxl2287: @eric-murray @hdamker I just investigated in our current configuration of using a code-generator.
I found the perfect configuration to fix also the allOf issue.

So I rollback my feedback here about maybe problematic `allOf`

RandyLevensalor (dismissed):
The device identifier documentation needs to be updated for qos-profile as well.  This was added to the other two profiles.

```    At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device. Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the session creation response.
      Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
```

Do we need to add some tests with multiple device identifiers?

code/API_definitions/quality-on-demand.yaml
jlurien: Thanks for the clarification, @eric-murray. As most people rely on viewers to understand the API, I'm OK to make it more readable. It is also important to ensure that we don't create unnecessary problems for code generation tools. If @maxl2287 has managed to make it work, though, I'm also fine with that.

jlurien (approved):
LGTM

RandyLevensalor:
> > The device identifier documentation needs to be updated for qos-profile as well. This was added to the other two profiles.
> > > Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
> 
> That text was already there in those two files - I just moved it to a standalone line to separate it from the definition of a device.

@eric-murray the qos-profile text isn't as clear as the new text.

It just says:
```Additionally, when a two-legged access token is used, the operation `retrieveQoSProfiles` allows the API consumer to optionally specify the target device in the request body.```

It doesn't mention what happens when multiple identifiers are used.

Could you add this

```Where more than one device identifier is provided, only one identifier will be selected by the implementation and this choice indicated to the API consumer in the session creation response.```

File: code/API_definitions/qos-profiles.yaml [hunk 3]
```diff
@@ -461,6 +466,19 @@ components:
...
           $ref: '#/components/schemas/QosProfileStatusEnum'
 
+    QosProfileDeviceResponse:
+      description: |
+        Response object for QoS Profiles for a given device. A device identifier is provided for disambiguation when multiple device identifiers were provided in the request.
+      allOf:
+        - type: object
+          properties:
+            device:
+              description: Disambiguates the device to which the session applies when multiple device identifiers were provided
```
hdamker, line 476: There is no "session" here. So at least the text need to be changed. 

But I would question if we really want to add the device to the QoSProfile objects. I would rather don't do the "device ambiguation" at all if there is no other option. The QoSProfile is more or less just a "product description", there shouldn't be a device property in.

```suggestion
              description: Disambiguates the device to which the session applies when multiple device identifiers were provided
```

hdamker (changes requested):
> * For `qos-profiles`, I included an optional `device` parameter in all profiles returned by the `/retrieve-qos-profiles` endpoint. This is clumsy, but has the advantage of being a non-breaking change (for clients). A better solution would be to have a separate `device` parameter common to all profiles returned, but this would be a breaking change.

On a second look I revoke my approval. The solution for qos-profile complicates the API in understandability and implementation just for the case of two-legged and a device object with multiple identifiers in a way which isn't logical (why should there be a device property in each QoSProfile only in this rare case, but in no other case).

Adding the DeviceResponse aside the array of QoSProfiles would be indeed a breaking change, as we haven't designed for extensibility (by defining a response object rather than the basic type directly). 

My proposal is to postpone the addition of the DeviceResponse object for this API until we have further reasons for a new major API version with proper extensibility.

code/API_definitions/qos-profiles.yaml
eric-murray: There are three alternative options that could be considered:
- If a device identifier needs to be returned for disambiguation, put it as a separate field before all the profiles. This would be a breaking change.
- Limit the number of device identifiers that can be specified (when filtering by device) to one only. This would also be a breaking change.
- Don't give any feedback as to which device identifier was selected if the API consumer specifies more than one. Either specify a "default order" in which they will be used if present (e.g. phone number first, then IP address if phone number not present, and so on), or just leave the API consumer assuming that their multiple identifiers did match the same device.

hdamker: For this API is would consider for now the third option with "just leave the API consumer assuming that their multiple identifiers did match the same device". The device will not be actually touched or changed by the API. If the API consumer then use the or a returned QoSProfile in one of the QoS APIs, they will get there the actual DeviceResponse (given that they use the same Device object also in this request).

There is even reason why it might be better to not return a DeviceResponse in this API: operators or aggregators who will just return a small set of static profiles might not even go to the actual network to resolve the device object -- with the DeviceResponse they would need to do so.

eric-murray:
For `qos-profiles`, the consequences of the API consumer and API provider having a different understanding of the `device` for which QoS profiles are being filtered are relatively minor - possibly the API consumer thinks they have a profile they can use, and then finds out they cannot when they request it, or possibly they do not find the profile they would have preferred to use, even though it was available to them.

The `DeviceResponse` schema was introduced to give a mechanism to disambiguate the `device` that is the subject of the API that avoids breaking changes in APIs. If we decide to postpone a fix until we have breaking changes to implement, probably the best solution is just to limit the number of device identifiers that can be specified by the API consumer to one only, which avoids any ambiguity.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -1437,15 +1451,37 @@ components:
...
+          ipv4Address: 198.51.100.0/24
+        qosProfile: QOS_L
+        sink: https://application-server.com/notifications
+        sessionId: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+        duration: 3600
+        qosStatus: REQUESTED
+
+    SESSION_CREATION_EXAMPLE_WITH_DEVICE_DISAMBIGUATION:
+      summary: Successful session creation with device disambiguation
+      description: Example response after a new QoS session has been created using multiple device identifiers
```
jlurien, line 1468:
```suggestion
      description: Example response after a new QoS session has been created when the API consumer included the `device` parameter in their request. Only one identifier is returned even if more than one device identifier was specified.
```

code/API_definitions/quality-on-demand.yaml
```diff
@@ -1437,15 +1451,37 @@ components:
...
+        applicationServer:
+          ipv4Address: 198.51.100.0/24
+        qosProfile: QOS_L
+        sink: https://application-server.com/notifications
+        sessionId: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+        duration: 3600
+        qosStatus: REQUESTED
+
+    SESSION_CREATION_EXAMPLE_WITH_DEVICE_DISAMBIGUATION:
+      summary: Successful session creation with device disambiguation
```
jlurien, line 1467:
```suggestion
      summary: Successful session creation with potential device disambiguation
```

jlurien (commented):
The current examples for disambiguation may be a bit misleading, as device may be returned also when only one device identifier is provided.

Reviewing Commonalities guidelines I think that current guidelines are not totally deterministic:

CAMARA-API-Design-Guide says "...Where an API consumer provides more than one device identifier, the API provider **should** indicate in the response which of these identifiers they are using to fulfil the API..."

CAMARA_common.yaml indicates: "... If the API consumer provides more than one device identifier in their request, the API provider **must** return a single identifier..."

and nothing is said for the case when only one identifier was provided in a 2-legged scenario, so we can assume that in this case the implementaion **may** return device in the response

eric-murray:
> My concern came from the "ambiguity" in the use of must/should in the guidelines, which I don't know it was intentional or because it was not discussed.

The intention was to provide the `DeviceResponse` schema for APIs that want to use it, strongly hint it should be used, but not mandate it. I think it is true that, if an API has adopted the `DeviceResponse` schema, then it must be used to resolve device ambiguities, but agree that the Design Guide does not state that.

If an API is really happy to allow an ambiguity to exist in the identification of the device when multiple device identifiers are provided, they are allowed to do that. The language of the Design Guide was chosen to draw developer's attention to the possibility of such ambiguity, strongly hint that it would be a good idea to resolve such ambiguity by responding with the device identifier used by the implementation, but not mandate it.

But, as Herbert notes above, we can just make the API consumer responsible to ensure all their identifiers derive from the same device, not tell them which device is assumed, and they must live with the consequences if the "wrong" device is used by the implementation.

eric-murray:
@camaraproject/quality-on-demand_maintainers
PR updated as discussed during meeting:
- `qos-profiles`
  - `DeviceResponse` object removed
  - Updated `info.description` to clarify that the API provider will decide which device identifier to use if multiple are provided
  - I left in one of the list examples as I think this is better than the OAS viewer auto-generated examples
- `quality-on-demand` and `qod-provisioning`
  - Updated `info.description` to clarify that the `device` may or may not be included in the response when only a single device identifier is provided
  - Updated example descriptions to be more generic regarding when they would apply

code/API_definitions/qod-provisioning.yaml
```diff
@@ -1080,9 +1100,29 @@ components:
...
+        qosProfile: QOS_L
+        sink: https://application-server.com/callback
+        provisioningId: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+        status: REQUESTED
+
     PROVISIONING_AVAILABLE:
-      summary: QoD provisioning status is available
-      description: The provisioning has become available
+      summary: QoD provisioning status is available with device in the provisioning details
+      description: The provisioning has become available, and the device to which the provisioning applies is included in the provisioning details
```
hdamker, line 1125:
```suggestion
      summary: QoS provisioning status is available with device in the assignment details
      description: The QoS profile assignment has become available, and the device to which the assignment applies is included in the assignment details
```

code/API_definitions/qod-provisioning.yaml
```diff
@@ -1119,7 +1154,7 @@ components:
         statusInfo: NETWORK_TERMINATED
 
     PROVISIONING_AVAILABLE_WITHOUT_DEVICE:
-      summary: QoD provisioning status is available but no device information is provided
+      summary: QoD provisioning status is available
```
hdamker, line 1157:
```suggestion
      summary: QoS provisioning status is available

```

hdamker (commented):
@eric-murray @jlurien I've tried to apply already the refined terminology from #447. That will not avoid (all) merge conflicts, but it should avoid that we oversee a necessary change in the cases where there is no merge conflict.

@jlurien would you check the suggestions and commit them?

hdamker:
> The device identifier documentation needs to be updated for qos-profile as well. This was added to the other two profiles.
> 
> ```
>       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
> ```

@RandyLevensalor within qos-profile there is the mandatory "Note2: ..." as in the other APIs in line https://github.com/camaraproject/QualityOnDemand/blob/34fb15ea2fb0221785ef77b3dadf0aaca3aa5c1c/code/API_definitions/qos-profiles.yaml#L476

If you want to have additional information as in quality-on-demand and qod/qos-provisioning in the info.description that would be a new issue/PR to update/extent the  info.description of qos-profiles.

Said this, it would be great if you can approve the PR here and unblock it from your change request.

hdamker (approved):
The PR LGTM already as is. The changes proposed in https://github.com/camaraproject/QualityOnDemand/pull/453#pullrequestreview-2969265330 can be either applied here or when resolving the merge conflicts with #477.

Important is that we get the PR here merged to be able to rebase #477.

jlurien (dismissed):
LGTM, with a suggestion

hdamker:
@eric-murray I'm no longer convinced that it have been a good idea to partial change the terminology here in this PR ;-)

eric-murray:
I updated all the examples to use "ASSIGNMENT" rather than "PROVISIONING", but the example contents get linted against the schema, and so changing `provisioningId` to "assignmentId" gave linting errors. Rather than change the schema, I left the example field as `provisioningId` for now.

jlurien:
I would merge this one first

jlurien (approved):

hdamker (approved):

hdamker:
@eric-murray @jlurien I've dismissed @RandyLevensalor older blocking review. I would be fine if you merge already, but also ok if you wait for a review and approval from him.

eric-murray:
> @eric-murray @jlurien I've dismissed @RandyLevensalor older blocking review. I would be fine if you merge already, but also ok if you wait for a review and approval from him.

I'll wait to the end of the day to give Randy a chance to review and make any comments on the recent changes

PR #460, opened 2025-06-23
Add reference booking Id for provisioning

RamTMO:
#### What type of PR is this?

* enhancement/feature

This PR allows the user to reference an already existing booking for provisioning the device

Fixes #

#### Changelog input

```
 release-note

1. Added BookingId Schema
2. Added bookingId property in BaseProvisioningInfo

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -374,10 +374,10 @@ components:
         example: "b4333c46-49c0-4f62-80d7-f0ef930f1c46"
 
   schemas:
-  
-  # This is an optional schema item, and is experimental and could be changed or removed in the future
+
+    # This is an optional schema item, and is experimental and could be changed or removed in the future
     BookingId:
-      description: | 
+      description: |
         ** Note** bookingId is optional and experimental and could be changed or removed in the future.
         If booking ID is present, provisioning of the device will happen against the booking done via QoSBooking API.
         If booking ID is present, requested qosProfile must match the qosProfile in the booking.
@@ -399,7 +399,7 @@ components:
         qosProfile:
           $ref: "#/components/schemas/QosProfileName"
         bookingId:
-          $ref: "#/components/schemas/BookingId"    
+          $ref: "#/components/schemas/BookingId"
         sink:
           type: string
           format: uri
--- a/code/API_definitions/qod-provisioning.yaml
+++ b/code/API_definitions/qod-provisioning.yaml
@@ -374,6 +374,17 @@ components:
         example: "b4333c46-49c0-4f62-80d7-f0ef930f1c46"
 
   schemas:
+  
+  # This is an optional schema item, and is experimental and could be changed or removed in the future
+    BookingId:
+      description: | 
+        ** Note** bookingId is optional and experimental and could be changed or removed in the future.
+        If booking ID is present, provisioning of the device will happen against the booking done via QoSBooking API.
+        If booking ID is present, requested qosProfile must match the qosProfile in the booking.
+        Booking Identifier in UUID format.
+      type: string
+      format: uuid
+
     ProvisioningId:
       description: Provisioning Identifier in UUID format
       type: string
@@ -387,6 +398,8 @@ components:
           $ref: "#/components/schemas/Device"
         qosProfile:
           $ref: "#/components/schemas/QosProfileName"
+        bookingId:
+          $ref: "#/components/schemas/BookingId"    
         sink:
           type: string
           format: uri


jlurien:
@RamTMO we should open an issue first and discuss the necessity for this there. I guess you are trying to relate this with QoS Booking but both APIs are in different stages and have different purposes. For example, in this API there are no areas or startTime, it's similar to quality-on-demand but with no limited duration.

code/API_definitions/qod-provisioning.yaml
```diff
@@ -387,6 +398,8 @@ components:
           $ref: "#/components/schemas/Device"
         qosProfile:
           $ref: "#/components/schemas/QosProfileName"
+        bookingId:
+          $ref: "#/components/schemas/BookingId"
```
tlohmar, line 402: What should happen, when the booked QOS profile (i.e. referred by the BookingId) is different than the QOS Profile here in this API call?
Should we add a new error code or make the QOS profile optional in QOD Provisioning?  

What should happen, when the QOD Provisioning API call is executed outside of the booking time window, e.g. before the booking start time or after the booking end time (i.e. after start time + booking duration)?

RamTMO: @jlurien Having a separate issue is a good idea. Yes, you are right this is related to QoS Booking Issue12. 

@tlohmar @jlurien  In this API, bookingId is optional.  I checked again, only qosProfile is required. The fundamental idea is, if bookingID is present, then this API gets the context around the bookingID which has Service Area, Service Duration, and Service Time etc.,  otherwise it does not have the booking context.   So, if bookingId is not present then there are no issues.  It has no context of booking, so it works on the on-demand execution.  

If the context is present (I mean bookingId) and if the qosProfile is different, or the call is made outside of the Service Window then it is an error condition.  Currently, we can piggy bag on existing errors to reduce complexities.  The rightway to do this: status = UNAVAILABLE and  StatusInfo = 'appropriate code'.  Thoughts?

RamTMO:
> /easycla

Still it is not working for this PR (it worked for other PR). I raised a support ticket.

hdamker:
@RamTMO Made the linting check happy by removing some trailing spaces and indent the comment.

@jlurien yes, issue and discussion make sense. My understanding is that this minimal invasive addition to allow to test qos-provisioning as the provisioning part of https://github.com/camaraproject/QoSBooking/issues/12, corresponding with https://github.com/camaraproject/QoSBooking/pull/23, see @RamTMO's [comment](https://github.com/camaraproject/QoSBooking/issues/12#issuecomment-2997691601).

@tlohmar valid questions - short-term existing error responses and status infos (e.g. NETWORK_TERMINATED) can be used, even if not a 100% fit. With some more experience if and how the APIs are fitting together some additional responses and status infos can be added (in a later version). BTW: the same questions came to my mind when reviewing dedicated-network-accesses in DedicatedNetworks, see my [comment](https://github.com/camaraproject/DedicatedNetworks/pull/34#issuecomment-3000734419) there.

RamTMO:
@camaraproject/quality-on-demand_maintainers Could you please review and support/approve this PR and the corresponting [PR#23 ](https://github.com/camaraproject/QoSBooking/pull/23)

code/API_definitions/qod-provisioning.yaml
jlurien: But both APIs have different scope and a totally different (non-compatible) model. Here a provisioningId is linked to the indefinite assignment of a required qosProfile to a required device, *nothing else*. In QoS Booking, the bookingId is linked to the booking of *session* (serviceArea, duration, startTime, device, qosProfile, applicationServer...). Indeed the model in QoS Booking is more similar to a session in QoD (but without area), so what do you mean when you say that this API gets the context from the bookingId for this API that only requires device+qosProfile. Moreover you are even proposing to make device optional for a booking, so what would be taken from the booking, the qosProfile?

jlurien:
> @jlurien yes, issue and discussion make sense. My understanding is that this minimal invasive addition to allow to test qos-provisioning as the provisioning part of [camaraproject/QoSBooking#12](https://github.com/camaraproject/QoSBooking/issues/12), corresponding with [camaraproject/QoSBooking#23](https://github.com/camaraproject/QoSBooking/pull/23), see @RamTMO's [comment](https://github.com/camaraproject/QoSBooking/issues/12#issuecomment-2997691601).

As I commented above, the data models of this API and QoS Booking are very different. This API cannot be used as the provisioning part of QoS Booking because this API is only intended for the permanent assignment of a qosProfile to a device. Concepts like duration or serviceArea, or the applicationServer part has no meaning here. The model of QoS Booking is much more related to quality-on-demand, but I don't see adding a bookingId there either.

So when I asked for an issue for this PR I mean an issue in this repo where we can properly discuss how a bookingId is useful here.

PR #522, opened 2026-01-05
[Added] Allow application servers to be identified by IP address list

### LINKED ISSUE #513: Allow more control over application server IP address specification
**Problem description**
Currently, the application server IP address(es) can be specified as a single IPv4 subnet and/or a single IPv6 subnet. For an established connection, where the application server IP address is already known, this is fine. Or if it intended to apply the QoS profiles to a whole subnet, this is also fine.

But if it is desired to create a QoS session that would apply to a connection once established between a device and a given application server, it is generally not possible to restrict the QoS session in advance to those IP addresses that the application server might use.

The device will be allocated the required application server IP address by DNS. It is normal for DNS entries to resolve to a small set of IP addresses (maybe 2 or 3), and DNS randomises the order in which these are returned to allow for load balancing.

So the possible set of IP addresses can be known in advance. But the QoS session can only be established in advance by specifying the minimum subnet that includes all possible IP addresses that the application server might use. This can be a very large subnet indeed.

**Possible evolution**
Allow more flexibility in specifying application server IP addresses by allowing an FQDN for application server to be specified, e.g.
```
  ...
  "applicationServer": {
    "fqdn": "example.com"
  },
  ...
```
This solution is non-breaking. In this case, all IP addresses returned by DNS for "example.com" would be included in the QoS session. Note that:
- the QoS profile would therefore apply to any application using these IP addresses, and not just `example.com`
- this solution also only works for application servers that have a public DNS entry.

**Alternative solution**
Alternatively, allow an array of IP subnets (IPv4 and/or IPv6) to be specified, e.g.
```
  ...
  "applicationServer": {
    "ipAddresses": ["2600:1406:5e00:6::17ce:bc1b","2600:1406:5e00:6::17ce:bc12","2600:1408:ec00:36::1736:7f31","2600:1408:ec00:36::1736:7f24","2600:1406:bc00:53::b81e:94ce","2600:1406:bc00:53::b81e:94c8","23.220.75.232/32","23.215.0.138/32","23.215.0.136/32","23.192.228.84/32","23.192.228.80/32","23.220.75.245/32"]
  },
  ...
```
This solution could be non-breaking (by keeping existing `ipv4Address` and `ipv6Address` options), or could be breaking by dropping these options.

**Additional context**

eric-murray:
#### What type of PR is this?
* enhancement/feature

Currently, application servers can only be identified by a contiguous set of IP addresses within a defined subnet. This PR allows application servers to be identified by a list of single IPv4 and IPv6 addresses.

Fixes #513 

None

#### Changelog input
```
 release-note
 - Allow application servers to be identified by IP address list
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -32,10 +32,8 @@ info:
 
     * **Identifier for the application server**:
     The domain name and/or IPv4 address(es) and/or IPv6 address(es) used by the application server (application backend). Domain names shall be expanded to a list of IP addresses using the network operator's own DNS service, and the QoS profile will apply to all IP packets with these IP addresses as source or destination address.
-    
+
     The application service provider should ensure, that the application clients use the network operator's own DNS service for resolving the application servers IP addresses. The application client should use the same application server IP address(es) as the network for a correction functioning of the feature.
-    
-    Note that application data flows of WebRTC (e.g. speech flow) cannot be described using FQDNs, as the IP addresses of the peer (e.g. a TURN server) is typically not based on a DNS resolution.
 
     * **App-Flow (between the application client and application server)**:
     The precise application data flow the developer wants to prioritize and have stable latency or throughput for. This flow is in the current API version determined by the identified device and the application server. And it can be further elaborated with details such as ports or port-ranges. Future version of the API might allow more detailed flow identification features.
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -878,7 +878,7 @@ components:
       description: |
         A server hosting backend applications to deliver some business logic to clients.
 
-        The developer must provide one or more of the following application server identifiers:
+        The API consumer must provide one or more of the following application server identifiers:
 
         * `domainName`
           - The domain name that the device may use to connect to the application server
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -912,6 +912,7 @@ components:
         oneOf:
           - format: ipv4
           - format: ipv6
+      uniqueItems: true
       minItems: 1
       maxItems: 16
 
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -31,7 +31,7 @@ info:
       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
     * **Identifier for the application server**:
-    IPv4 and/or IPv6 address of the application server (application backend)
+    The domain name and/or IPv4 address(es) and/or IPv6 address(es) used by the application server (application backend). Domain names shall be expanded to a list of IP addresses using the network operator's own DNS service, and the QoS profile will apply to all IP addresses specified.
 
     * **App-Flow (between the application client and application server)**:
     The precise application data flow the developer wants to prioritize and have stable latency or throughput for. This flow is in the current API version determined by the identified device and the application server. And it can be further elaborated with details such as ports or port-ranges. Future version of the API might allow more detailed flow identification features.
@@ -876,12 +876,20 @@ components:
       description: |
         A server hosting backend applications to deliver some business logic to clients.
 
-        The developer can choose to provide the below specified device identifiers:
+        The developer must provide one or more of the following application server identifiers:
 
+        * `domainName`
         * `ipv4Address`
         * `ipv6Address`
+        
+        If more than one identifier is provided, the QoS profile shall apply to all application servers identified.
+
+        If the developer specifies a domain name, the API provider shall resolve that domain name to a list of IP addresses using the DNS service that the network operator provides for devices connected to their network. The QoS profile shall apply to all traffic using these IP addresses, which may include traffic for a different domain name that shares one or more of these IP addresses. Note that if the device itself uses a different DNS service, there is no guarantee that the QoS profile will apply to the IP addresses returned by that service for the specified domain name.
+        
       type: object
       properties:
+        domainName:
+          $ref: "#/components/schemas/ApplicationServerDomainName"
         ipv4Address:
           $ref: "#/components/schemas/ApplicationServerIpv4Address"
         ipv6Address:
@@ -939,6 +947,13 @@ components:
       type: string
       format: ipv6
       example: 2001:db8:85a3:8d3:1319:8a2e:370:7344
+
+    ApplicationServerDomainName:
+      type: string
+      format: hostname
+      example: "example.com"
+      description: |
+        The domain name used by the application server. The domain name must be resolvable using the DNS service provided by the network operator for devices connected to their network. It is expected that the domain name will usually be a fully qualified domain name (FQDN).
 
     ApplicationServerIpv4Address:
       type: string
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -879,8 +879,13 @@ components:
         The developer must provide one or more of the following application server identifiers:
 
         * `domainName`
+          - The domain name that the device may use to connect to the application server
+        * `ipAddresses`
+          - A list of IPv4 and/or IPv6 addresses that may be used by the application server
         * `ipv4Address`
+          - An IPv4 subnet
         * `ipv6Address`
+          - An IPv6 subnet
 
         If more than one identifier is provided, the QoS profile shall apply to all application servers identified.
 
@@ -890,11 +895,25 @@ components:
       properties:
         domainName:
           $ref: "#/components/schemas/ApplicationServerDomainName"
+        ipAddresses:
+          $ref: "#/components/schemas/ApplicationServerIPAddressList"
         ipv4Address:
           $ref: "#/components/schemas/ApplicationServerIpv4Address"
         ipv6Address:
           $ref: "#/components/schemas/ApplicationServerIpv6Address"
       minProperties: 1
+
+    ApplicationServerIPAddressList:
+          description: A list of IPv4 and/or IPv6 addresses that may be used by the application server
+          type: array
+          items:
+            description: An IP address that may be used by the application server
+            type: string
+            oneOf:
+              - format: ipv4
+              - format: ipv6
+          minItems: 1
+          maxItems: 16
 
     NetworkAccessIdentifier:
       description: A public identifier addressing a subscription in a mobile network. In 3GPP terminology, it corresponds to the GPSI formatted with the External Identifier ({Local Identifier}@{Domain Identifier}). Unlike the telephone number, the network access identifier is not subjected to portability ruling in force, and is individually managed by each operator.
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -916,7 +916,7 @@ components:
       items:
         description: An IP address that may be used by the application server
         type: string
-        oneOf:
+        anyOf:
           - format: ipv4
           - format: ipv6
       uniqueItems: true
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -881,11 +881,11 @@ components:
         * `domainName`
         * `ipv4Address`
         * `ipv6Address`
-        
+
         If more than one identifier is provided, the QoS profile shall apply to all application servers identified.
 
         If the developer specifies a domain name, the API provider shall resolve that domain name to a list of IP addresses using the DNS service that the network operator provides for devices connected to their network. The QoS profile shall apply to all traffic using these IP addresses, which may include traffic for a different domain name that shares one or more of these IP addresses. Note that if the device itself uses a different DNS service, there is no guarantee that the QoS profile will apply to the IP addresses returned by that service for the specified domain name.
-        
+
       type: object
       properties:
         domainName:
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -31,7 +31,11 @@ info:
       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
     * **Identifier for the application server**:
-    The domain name and/or IPv4 address(es) and/or IPv6 address(es) used by the application server (application backend). Domain names shall be expanded to a list of IP addresses using the network operator's own DNS service, and the QoS profile will apply to all IP addresses specified.
+    The domain name and/or IPv4 address(es) and/or IPv6 address(es) used by the application server (application backend). Domain names shall be expanded to a list of IP addresses using the network operator's own DNS service, and the QoS profile will apply to all IP packets with these IP addresses as source or destination address.
+    
+    The application service provider should ensure, that the application clients use the network operator's own DNS service for resolving the application servers IP addresses. The application client should use the same application server IP address(es) as the network for a correction functioning of the feature.
+    
+    Note that application data flows of WebRTC (e.g. speech flow) cannot be described using FQDNs, as the IP addresses of the peer (e.g. a TURN server) is typically not based on a DNS resolution.
 
     * **App-Flow (between the application client and application server)**:
     The precise application data flow the developer wants to prioritize and have stable latency or throughput for. This flow is in the current API version determined by the identified device and the application server. And it can be further elaborated with details such as ports or port-ranges. Future version of the API might allow more detailed flow identification features.
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -904,16 +904,16 @@ components:
       minProperties: 1
 
     ApplicationServerIPAddressList:
-          description: A list of IPv4 and/or IPv6 addresses that may be used by the application server
-          type: array
-          items:
-            description: An IP address that may be used by the application server
-            type: string
-            oneOf:
-              - format: ipv4
-              - format: ipv6
-          minItems: 1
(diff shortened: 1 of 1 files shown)


code/API_definitions/quality-on-demand.yaml
```diff
@@ -876,12 +876,20 @@ components:
...
-        The developer can choose to provide the below specified device identifiers:
+        The developer must provide one or more of the following application server identifiers:
 
+        * `domainName`
         * `ipv4Address`
         * `ipv6Address`
+
+        If more than one identifier is provided, the QoS profile shall apply to all application servers identified.
+
+        If the developer specifies a domain name, the API provider shall resolve that domain name to a list of IP addresses using the DNS service that the network operator provides for devices connected to their network. The QoS profile shall apply to all traffic using these IP addresses, which may include traffic for a different domain name that shares one or more of these IP addresses. Note that if the device itself uses a different DNS service, there is no guarantee that the QoS profile will apply to the IP addresses returned by that service for the specified domain name.
```
tlohmar:
```suggestion
        If the API Invoker specifies a domain name, the API provider shall resolve that domain name to a list of IP addresses using the DNS service that the network operator provides for devices connected to their network. The QoS profile shall apply to all traffic using these IP addresses, which may include traffic for a different domain name that shares one or more of these IP addresses. Note that if the Application Client on the device uses a different DNS service, there is no guarantee that the QoS profile will apply to the IP addresses returned by that service for the specified domain name.
```

tlohmar: Aha, I have not checked the used terms. Adjusted now.

When checking, I found that two terms are used, i.e. API consumer and Application developer (or just developer). I assume that the term "application developer" refers to the developer of the API consumer, i.e. Application backend developer.

ilya-smirnov-berlin:
I find this suggested extension reasonable but have one remark.

If an API consumer has the ability to resolve a given domain name into a list of IP addresses automatically on the provider side and apply all of them to QoS flows, then the API should also allow the consumer to create a session using such an IP address list for the application server, rather than being limited to a single single network segment (single IP in particular case). For example:

```
  "applicationServer": {
    "ipv4Addresses": [ "98.138.91.63", "72.30.3.157", "66.163.190.174" ]
  }
```

As already mentioned, DNS resolution results may differ from the consumer’s own resolution, even when using the same DNS server, due to region-based responses and load balancing. This behavior should be explicitly documented as warning to consumers.

eric-murray:
> Maybe we have to remark also that the applicationServer in the response must be the same indicated in the request, without explicit translation to the IP addresses.

For this PR, yes - indeed, it will usually not be possible to express the IP addresses associated a domain name using the current schema based on a single IPv6 and/or IPv6 subnet.

But I will separately propose a breaking change whereby the session creation response includes a list (i.e. array) of all application server IP addresses included in the QoS session, rather than the current `ip4Address` and `ipv6Address` parameters. That can be discussed separately to this PR, as it can be an independent change.

jlurien (commented):
It would make sense to remove support for subnets with the new property for IP addresses (list), but how would we cover the case where API consumers are filling 0.0.0.0/0 as `applicationServer`, meaning any destination? We could make `applicationServer` optional for this case.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -876,18 +878,46 @@ components:
       description: |
         A server hosting backend applications to deliver some business logic to clients.
 
-        The developer can choose to provide the below specified device identifiers:
+        The developer must provide one or more of the following application server identifiers:
```
tlohmar:
```suggestion
        The API consumer must provide one or more of the following application server identifiers:
```

tlohmar (commented):
There are seven other occurrences of "developer" in the yaml. Are these worth a separate PR (editorial changes) or ok to fix those in this PR?

eric-murray:
> It would make sense to remove support for subnets with the new property for IP addresses (list), but how would we cover the case where API consumers are filling 0.0.0.0/0 as `applicationServer`, meaning any destination? We could make `applicationServer` optional for this case.

Hi @jlurien 

So this PR would not affect the ability to specify /0 subnets, as the existing properties `ipv4Address` and `ipv6Address` would remain. Hence /0 subnets could continue to be specified as they are currently if this PR was merged as is.

When considering issue #528, it may be that we agree to make subnets more restricted, or perhaps even remove the ability to specify subnets completely. That is my current preference, as /0 subnets are anyway implicitly supported by the qos-provisioning API, so anyone who wants a /0 subnet should use that API. But if we want to keep /0 as a special case for this API, I agree making `applicationServer` optional would be a good solution.

So the future of /0 subnets will be discussed as part of #528.

eric-murray:
> There are seven other occurrences of "developer" in the yaml. Are these worth a separate PR (editorial changes) or ok to fix those in this PR?

Once #547 is merged, I will fix any conflicts with this PR, maintaining any changes in #547.

eric-murray:
Conflicts following merge of #547 now resolved

jlurien:
Hi @eric-murray,

Regarding the **removal of subnets** (masks), it seems that they are currently used in some of our integrations, and our Product team is in favor of keeping them to support existing use cases. But as you say this is out of the scope of this PR, so we can leave this discussion for another ocassion.

Focusing on the **domain as identifiers**, I received some internal concerns regarding "domainName" as a flow identifier: 

From a network enforcement perspective, QoS works on IP-level attributes (IPs, ports) and a domain name is just an external DNS step that isn't deterministic.
The main issue is that DNS resolution can be different for the app and the API provider. For example:

1. The client app resolves "europe.netflixserver.net" and gets 10.10.10.40. (Netflix CDN operates under multiple FQDNs with global pools of more than 18.000 IPs)
2. Client app request QoD targetting the FQDN and the operator resolves the same FQDN but gets 10.10.10.50 due to load balancing or CDN steering.
3. The network creates a dedicated bearer for .50, but the actual video traffic flows to .40.
4. Result: The API returns a "Success" 201 code, but the user gets zero QoS because the traffic doesn't match the filter. This is a silent failure and hard to debug.

Adding all the IPs associated with that domain to the QoS session isn't technically feasible. For a service like Netflix (for example), they have thousands of edge IPs (Open Connect CDN). Even for a single FQDN, you might get dozens of rotating IPs..

Using explicit IP addresses is the only way to ensure the QoS actually hits the right traffic. If we move forward with "domainName", we need to know how these mismatches and the "IP explosion" are handled. Otherwise, operators should be able to reject FQDN-based requests or we should postpone this until there's a more robust solution.

eric-murray:
Hi @jlurien 

So, if I understand the problem correctly, some service providers who provide their services via CDN are configuring public DNS lookups to be dynamic, such that repeated lookups for, say, `europe.netflixserver.net` result in a different set of IP addresses each time?

I'm not familiar enough with how CDNs configure their DNS lookups to challenge this, but the primary intended use case for my proposed change is to allow the QoS session to be created in advance of actually establishing the flow with the application server, and hence at a point of time when the actual destination IP address is not known. A related secondary use case is where the SDK calling the API does not explicitly know the application server IP address being used, even after the flow is established.

For CDNs, it appears that the flow must be established first, before the QoS session is created, and hence the application server IP address will already be known and must be known for the QoS session to be created.

For cases where it is known that the FQDN is not deterministic, I'd propose the API provider return an error if that `domainName` is requested. Maybe `422 SERVICE_NOT_APPLICABLE`, or some custom code if we want to reserve `SERVICE_NOT_APPLICABLE` to cases where the service is not applicable to the indicated device.

In addition, as part of a PR associated with #529, I will be proposing that the response to the API consumer include an explicit list of the application server IP addresses that will be prioritised. So rather than just echoing the `domainName`, the actual IP addresses determined by the API provider will be returned.

If the API consumer ends up using an address for the application server that is not in that list, the QoS session can be deleted and a new one for the address actually being used created instead.

Hopefully these two proposed mitigations will ease your colleagues concern.

tlohmar:
> I'm not familiar enough with how CDNs configure their DNS lookups to challenge this, ...

CDNs relies typically on DNS Canonical Name (CNAME) entries for redirecting lets say `europe.netflixserver.net` to say `nnn.cdnprovider.com`. 

The DNS of the ASP does this first redirect to a Canonical Domain belonging to the CDN Provider. After that, the DNS of the CDN Provider takes over, and redirects either  from `nnn.cdnprovider.com` to `yyy.cdnprovider.com`, or from  `nnn.cdnprovider.com` to `1.2.3.4` (actual IP address)

Note, also Geo DNS principle may be involved.

The CDN Server is presenting the HTTPS certificate of the ASP, so that the HTTP Client can verify the server as authorized server. 

Thus, the Application Client must first connect to the CDN Server. During this process, the serving CDN Server is selected and the IP address of the selected CDN Server needs to be used for the QoS Session request.

eric-murray:
Hi @tlohmar 

Your explanation does not explain why the resulting list of IP addresses is dynamic.

Is that because the redirection to other domain names is dynamic (i.e. `europe.netflixserver.net` -> `nnn.cdnprovider.com` results in a non-deterministic and dynamically chosen `nnn`), or because the resolution from  `nnn.cdnprovider.com` to the actual IP addresses is itself dynamic, resulting in a different set each time?

hdamker:
@eric-murray the feedback from our engineering team was also that they won't be able to support the feature. The main reason is that the risk that they get a different DNS resolution than the device can't be mitigated on their side. How upstream DNS servers will resolve a request based on which criteria is out of our control.

tlohmar:
> Hi @tlohmar
>
> Your explanation does not explain why the resulting list of IP addresses is dynamic.
>
> Is that because the redirection to other domain names is dynamic (i.e. europe.netflixserver.net -> nnn.cdnprovider.com results in a non-deterministic and dynamically chosen nnn), or because the resolution from nnn.cdnprovider.com to the actual IP addresses is itself dynamic, resulting in a different set each time?

There are several reasons, why they may appear as dynamic:
A: Usage of GeoDNS - here, the DNS response is created based on the geographic (or topology) location of the requester. This may happen when redirecting from the ASP domain to the CDN Provider domain or when further redirecting within a CDN
B: CDN Provider selection - some ASP have a multi-CDN provider strategy, e.g. selecting a different CDN Provider (dynamically) based on region or for stability. For example, the ASP may redirect to a different CDN provider based on current CDN performance.
C: Load-Balancing / Load Management of CDN Server capacity - this is typically an action within a CDN, where the CDN provider is redirecting based on CDN Server load. Popular content is typically hosted at a higher number of CDN servers than long-tail content and the CDN provider often tries to keep the load on individual CDN Server equal.
D: I have likely missed some reasons.

Thus, it is a pretty deterministic process from within the ASP / CDN Provider. However, it looks random / dynamic from the outside.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -878,17 +882,45 @@ components:
...
+        * or by a single IPv4 subnet and/or single IPv6 subnet
+          - `ipv4Address`
+          - `ipv6Address`
+
+        If more than one identifier is provided, the QoS profile shall apply to all application servers identified.
+
+      oneOf:
+        - type: object
+          required:
+          - ipAddresses
```
hdamker:
```suggestion
            - ipAddresses
```

eric-murray:
Removed `domainName` proposal, but retained a list of IP addresses as an alternative to IP subnets (either one or the other).

Ready for review

code/API_definitions/quality-on-demand.yaml
```diff
@@ -878,17 +891,54 @@ components:
...
+      description: A list of IPv4 and/or IPv6 addresses that may be used by the application server
+      type: array
+      minItems: 1
+      maxItems: 16
+      items:
+        description: A single IP address that may be used by the application server
+        uniqueItems: true
+        anyOf:
+          - $ref: "#/components/schemas/SingleIpv4Addr"
+          - $ref: "#/components/schemas/SingleIpv6Addr"
```
hdamker, line 941: `uniqueItems: true` must be a field of the array, same level as `minItems` and `maxItems`.

```suggestion
    ApplicationServerIpAddressListType:
      description: A list of unique IPv4 and/or IPv6 addresses that may be used by the application server
      type: array
      minItems: 1
      maxItems: 16
      uniqueItems: true
      items:
        description: A single IP address that may be used by the application server
        anyOf:
          - $ref: "#/components/schemas/SingleIpv4Addr"
          - $ref: "#/components/schemas/SingleIpv6Addr"
```

hdamker (changes requested):
Looks good to me, but one fix to be made.

Further observations:
* Test definitions do not cover the new `ipAddresses` branch. There is currently only a generic “valid application server” wording. I propose to scope this as an own issue and not include here in the PR
* The PR title still says “identified by domain name" ... worth an update

code/API_definitions/quality-on-demand.yaml
eric-murray: Good catch. I added some "positive" examples to ensure that spectral agreed they were compliant with the updated schema, but adding "negative" examples that should not be valid is not so straightforward, as they need to be removed and then re-added any time it is required to confirm that they are still not valid.

code/API_definitions/quality-on-demand.yaml
```diff
@@ -878,17 +891,54 @@ components:
...
-        * `ipv4Address`
-        * `ipv6Address`
+        * by a list of individual IPv4 and/or IPv6 addresses:
+          - `ipAddresses`
+
+        * or by a single IPv4 subnet and/or single IPv6 subnet
+          - `ipv4Address`
+          - `ipv6Address`
+
+        If multiple IP addresses are specified, the QoS profile shall apply to all application servers using any of these addresses by the identified device
```
hdamker: Maybe a clearer replacement:

```suggestion
        If `ipAddresses` is specified, the QoS profile shall apply to traffic between the identified device and any of the listed application server IP addresses.
```

hdamker (dismissed):
LGTM now, just one non-blocking suggestion

hdamker (dismissed):
LGTM - thanks @eric-murray !

code/API_definitions/quality-on-demand.yaml
```diff
@@ -28,7 +28,7 @@ info:
       Note: Network Access Identifier is defined for future use and will not be supported with this version of the API.
 
     * **Identifier for the application server**:
-    IPv4 and/or IPv6 address of the application server (application backend)
+    A list or subnet of IPv4 and/or IPv6 addresses used by the application server (application backend)
```
RandyLevensalor: This wording is a bit awkward.

I think it be be a list of one or more ip addresses or a list of one or more subnets.

The subnet attribute is `ApplicationServerIpAddressSubnets` not a singular subnet.

RandyLevensalor (changes requested):
Align the description with the plurality of the schema names.

code/API_definitions/quality-on-demand.yaml
eric-murray: Well, it could be two (and a maximum of two) subnets - one IPv4 subnet and one IPv6 subnet. So whilst two subnets could be specified, I've always thought that one of those would be redundant for the resulting QoS session itself.

However, to capture the most general case, how about:
`A set of IPv4 and/or IPv6 addresses or subnets used by the application server (application backend)`

That captures the possibility that a plurality of subnets may be specified.

RandyLevensalor: @eric-murray that sounds better to me. Thanks!!

jlurien:
Removing `domainName` solves the DNS non-determinism issue, which was the main concern. Still, the `ipAddresses` list introduces additional complexity that makes more evident some issues that we already had:

- `applicationServerPorts` is kept as it was, and it applies to all IPs in `applicationServer`. So if the consumer sends up to 16 IPs and a few ports or port ranges, the implementation has to install many packet filters (IPs × ports) on the same QoS flow. This number may exceed what the implementation supports.
- The list can also contain IPs that the implementation cannot enforce in a single session.

If the implementation has to reject the request, the current 422 errors (`SERVICE_NOT_APPLICABLE`, `UNSUPPORTED_IDENTIFIER`, ...) do not clearly cover the case "the combination of `applicationServer` and `applicationServerPorts` cannot be supported". I think that the spec — here and also in the work coming from issue #577 — should make clear which error code to return when the implementation cannot create a session for the requested destination and ports.

hdamker (approved):
Just renewing my approval after the last editorial change.

RandyLevensalor (approved):
@eric-murray great work.  Thanks for making the update.

hdamker:
> If the implementation has to reject the request, the current 422 errors (`SERVICE_NOT_APPLICABLE`, `UNSUPPORTED_IDENTIFIER`, ...)

An API specific 422 error code is needed. Proposals?
Todo: create new issue for the above comment and defer it to a new PR.

jlurien:
LGTM, we can decide on the new code on a separate PR

jlurien (approved):

eric-murray:
@jlurien 

>applicationServerPorts is kept as it was, and it applies to all IPs in applicationServer. So if the consumer sends up to 16 IPs and a few ports or port ranges, the implementation has to install many packet filters (IPs × ports) on the same QoS flow. This number may exceed what the implementation supports.

We anyway need to limit the number of ports or port ranges for compliance with the updated design guidelines, but this has not been discussed. My preference is to greatly limit the ability to specify the application server ports - I would propose 1 single port OR 1 port range can be specified, but no more. Of course, this would considered a breaking change ...

Anyway, if you do not include this in the scope of your issue for an error code when the API provider cannot support the complexity of the specified flow rules, I will create a separate issue.

> The list can also contain IPs that the implementation cannot enforce in a single session.

This comment isn't clear to me. Is the issue that 16 separate IP addresses (requiring 16 separate flow rules) might be too many?

hdamker:
We are good to merge here

@jlurien @eric-murray  don't forget to create the discussed new issues.

PR #562, opened 2026-05-29
chore: prepare QoD release plan for rc

hdamker:
#### What type of PR is this?

subproject management

Update release-plan.yaml for the QoD release candidate preparation.

This sets the release type to pre-release-rc as agreed in CQM Call on 2026-05-29, moves the three API target statuses to rc, and updates the Commonalities dependency to r4.3.

N/A - no upstream issue.

N/A

#### Changelog input

```
release-note

NONE
```

#### Additional documentation

This section can be blank.

```
docs

NONE
```

### DIFF
--- a/release-plan.yaml
+++ b/release-plan.yaml
@@ -20,12 +20,12 @@ repository:
 
   # Release type being prepared (must be set before release can be triggered)
   # Options: none | pre-release-alpha | pre-release-rc | public-release | maintenance-release
-  target_release_type: none
+  target_release_type: pre-release-rc
 
 # Dependencies on Commonalities and ICM releases
 # Update per ReleaseManagement requirements for each release cycle
 dependencies:
-  commonalities_release: r4.1
+  commonalities_release: r4.3
   identity_consent_management_release: r4.1
 
 # APIs in this repository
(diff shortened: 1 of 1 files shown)


hdamker:
@eric-murray @jlurien @RandyLevensalor  Would be good if one of you can approve this PR timely - it will enable me to prepare the alignment with Commonalities r4.3. The plan just declares out intent that the next release will be a pre-release-rc within the Sync26 meta-release cycle. It does not define when we are actually ready to do this release (but there is the expectation from the meta-release time plan that this happens until end of June).

cc: @benhepworth

release-plan.yaml
RandyLevensalor, line 19: I think that I missed something.  Should the be changed from r4.1 to r4.2 since there are changes (like updating commonalities since the previous target in this release cycle?  Or is that just for updates after the Sync 26 release has been completed.

hdamker: Hi @RandyLevensalor: the last release of the repository was r3.2 (the public release of Fall25). We are starting now a new release cycle for Sync26, therefore is the release *number* r4.1 for this first pre-release in the cycle correct. The indication of changes will be mainly done by the API version (see lower part of the release-plan.yaml).

See also: https://github.com/camaraproject/QualityOnDemand/releases

RandyLevensalor (approved):
Thanks for the clarification.

PR #576, opened 2026-06-16
docs: clarify create-time QoD unavailability

### LINKED ISSUE #520: Clarify behaviour when a create request cannot be provided by the network
**Problem description**
The current specification for Quality on Demand (and also for QoS Provisioning and QoS Booking modeled after it), suggests that when the implementation determines that the request to create a session/ cannot be fulfilled, the answer should be  201, "UNAVAILABLE - The requested QoS cannot be provided by the network due to some reason", possible with statusInfo to NETWORK_TERMINATED.

But using 2xx for cases when the network refuses to fulfill the request is debatable. For example, when the implementation determines that the maximum capacity is reached and no more sessions or booking can be accepted. Should a resource be created and an ID assigned to it? Or should an error be returned instead?

**Expected action**
Align and clarify better the documentation for this scenario. 

**Additional context**
It applies also to QoS Booking, so a dedicated issue may be created there.

hdamker:
#### What type of PR is this?

documentation

This PR clarifies that `createSession` can return `201` with `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED` when the network determines during session creation that the requested QoS cannot be provided.

It also broadens the `NETWORK_TERMINATED` `statusInfo` description to cover that create-time non-fulfilment case without adding a new enum value.

Fixes #520

Doc-only clarification. No new `statusInfo` enum value is introduced, so the change remains suitable for `quality-on-demand` v1.2.0.

The existing `createSession` Gherkin test definition already allows this outcome at the generic success-scenario level: it expects a `201` response that complies with `SessionInfo`, and `statusInfo` is present only when `qosStatus` is `UNAVAILABLE`. There is no dedicated scenario that forces the create-time `UNAVAILABLE` branch.

#### Changelog input

```
release-note
Clarify that createSession may return qosStatus UNAVAILABLE with statusInfo NETWORK_TERMINATED when the network cannot provide the requested QoS during session creation.
```

#### Additional documentation

This section can be blank.

```
docs
```

### DIFF
--- a/code/API_definitions/quality-on-demand.yaml
+++ b/code/API_definitions/quality-on-demand.yaml
@@ -133,7 +133,10 @@ paths:
         - a `QOS_STATUS_CHANGED` event notification with `qosStatus` as `AVAILABLE` after the network notifies that it has created the requested session, or
         - a `QOS_STATUS_CHANGED` event notification with `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED` after the network notifies that it has failed to provide the requested session.
 
-        A `QOS_STATUS_CHANGED` event notification with `qosStatus` as `UNAVAILABLE` will also be send if the network terminates the session before the requested duration expired
+        If the network already knows during session creation that it cannot provide the requested QoS, the `201` response MAY directly include
+        `qosStatus` as `UNAVAILABLE` and `statusInfo` as `NETWORK_TERMINATED`.
+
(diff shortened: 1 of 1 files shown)


hdamker:
@eric-murray @jlurien @RandyLevensalor Any review comments?

RandyLevensalor (approved):

eric-murray (approved):

hdamker:
Merging to be able to go forward. 

@jlurien if you have comments or see needed corrections please raise them as new issue.

PR #591, opened 2026-07-31
fix(qos-profiles): give the two 404 responses non-conflicting names

### LINKED ISSUE #589: qos-profiles: Generic404 name collision produces Generic404-2 in the bundled API definition
**Problem description**

`qos-profiles.yaml` defines two different response bodies competing for the name `Generic404`:

| Component | Body | Code set | Used by |
|---|---|---|---|
| `Generic404` | full local definition | `NOT_FOUND` | `getQosProfile` |
| `GenericDevice404` | alias `$ref` → `../common/CAMARA_common.yaml#/components/responses/Generic404` | `NOT_FOUND`, `IDENTIFIER_NOT_FOUND` | `retrieveQoSProfiles` |

At release time the bundler must inline the common `Generic404` under its own name, but the
local code-subset already holds it. It therefore emits the inlined response as `Generic404-2`,
rewrites the `retrieveQoSProfiles` use-site directly to that name, and leaves `GenericDevice404`
unreferenced.

Two consequences:

* The published bundled API definition contains a component named `Generic404-2`, which
  propagates into generated client models.
* CAMARA Validation reports `S-211` (potentially unused component) on `GenericDevice404` — a real
  finding, not a false positive. It appeared as the only warning on the r4.1 release review.

Both code sets are substantively correct: `getQosProfile` takes only `name` and `x-correlator`,
so `IDENTIFIER_NOT_FOUND` cannot occur there. Only the naming is wrong.

`Generic404-2` is the only collision-renamed component across the three bundled specs in this
repository. Every other aliased common component in `qos-profiles.yaml` (`Device`, `Generic400`,
`Generic401`) keeps the common name locally and inlines in place without a suffix.

**Expected behavior**

Free the reserved name so the alias resolves cleanly:

* rename the local code-subset `Generic404` to an API-specific name, e.g. `QosProfileNotFound404`
* rename the alias `GenericDevice404` to `Generic404`, matching the `Generic400` / `Generic401`
  pattern already used in this file
* update the two use-sites (`getQosProfile` 404, `retrieveQoSProfiles` 404)

**Alternative solution**

The broader alignment is to drop the pure-alias indirection in `qos-profiles.yaml` altogether and
`$ref` the common file directly at every use-site, as `quality-on-demand.yaml` already does.
`qos-profiles.yaml` currently carries four alias components (`Device`, `Generic400`, `Generic401`,
`GenericDevice404`) where `quality-on-demand.yaml` has only `openId`, which must stay local because
the operation-level security requirement names it. Removing the indirection eliminates this class
of collision rather than renaming around one instance.

That alignment cannot be completed yet. The `Generic*` responses bundle error codes that do not
apply to every API, so the code-subset responses still have to be defined locally. The minimal
referenceable responses that would replace them — `NotFound404`, `PermissionDenied403`,
`DeviceIdentifier422` and the rest of the catalogue added in
https://github.com/camaraproject/Commonalities/pull/665 — become available with Commonalities r4.4.
Until this repository moves to r4.4, dedicated local error-response definitions remain necessary.

Suggested sequencing: apply the rename now, and revisit the direct-reference alignment when the
`commonalities_release` dependency moves to r4.4.

**Additional context**

The collision was introduced when `GenericDevice404` was converted from a local definition into an
alias of the common `Generic404`. In r3.2 both `Generic404` and `GenericDevice404` were distinct
local definitions with no shared name, and the specs were not bundled, so neither the
`Generic404-2` artifact nor the `S-211` finding occurred.

**Validation finding**

Reported on the r4.1 release review, https://github.com/camaraproject/QualityOnDemand/pull/588
([run](https://github.com/camaraproject/QualityOnDemand/actions/runs/30660683772)):

```
rule   file                                     line  level  message
S-211  code/API_definitions/qos-profiles.yaml   715   warn   Potentially unused component has been detected.
```

Run summary: `Passed: 6 findings (0 errors, 1 warnings, 5 hints)` — this is the only warning.

Status: **not deferred**. The fix is intended to land on `main` before the release; the current
snapshot is then discarded and re-created, so that the last release candidate and the public release
are content-identical.

hdamker:
#### What type of PR is this?

bug

`qos-profiles.yaml` had two different response bodies competing for the name `Generic404`: a local `NOT_FOUND`-only definition used by `getQosProfile`, and a `GenericDevice404` component that was a pure alias to the common `Generic404` (`NOT_FOUND` + `IDENTIFIER_NOT_FOUND`) used by `retrieveQoSProfiles`. Bundling must inline the common response under its own name, but the local definition held it, so redocly emitted a `Generic404-2` component, repointed the `retrieveQoSProfiles` use-site at it, and left `GenericDevice404` orphaned — surfacing as an unused-component warning on the bundled spec. This gives each 404 a non-conflicting, self-describing name and defines both locally: `NotFound404` for `getQosProfile` and `RetrieveQosProfilesNotFound404` for `retrieveQoSProfiles`.

* No `Generic404-2` in the bundled spec, and the unused-component warning is gone.
* No dependency on `Generic404`, which is deprecated in Commonalities 0.9.0 and slated for removal.
* The wire contract is unchanged: `retrieveQoSProfiles` still documents `NOT_FOUND` + `IDENTIFIER_NOT_FOUND`, `getQosProfile` still `NOT_FOUND` only.

Fixes #589

The warning only exists after bundling — in the source spec the alias is genuinely referenced — so it appears on release-review runs rather than ordinary branch runs. To reproduce, bundle `code/API_definitions/qos-profiles.yaml` with `redocly bundle` before and after this change and lint both outputs with `.spectral-r4.yaml`.

`RetrieveQosProfilesNotFound404` deliberately avoids any name in the Commonalities catalogue, whose 404s are `NotFound404` (`NOT_FOUND`) and `IdentifierNotFound404` (`IDENTIFIER_NOT_FOUND`) separately. Reusing a catalogue name for a different code set is what caused this collision.

Whether `retrieveQoSProfiles` should document `NOT_FOUND` at all is a separate question, not touched here.

#### Changelog input

```
 release-note
Fix a 404 response component name collision in qos-profiles introduced by #575; not present in any release.
```

#### Additional documentation

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -140,7 +140,7 @@ paths:
         "403":
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/GenericDevice404"
+          $ref: "#/components/responses/RetrieveQosProfilesNotFound404"
         "422":
           $ref: "#/components/responses/Generic422"
         "429":
@@ -185,7 +185,7 @@ paths:
         "403":
           $ref: "#/components/responses/Generic403"
         "404":
-          $ref: "#/components/responses/Generic404"
+          $ref: "#/components/responses/NotFound404"
         "429":
           $ref: "#/components/responses/Generic429"
 
@@ -545,7 +545,7 @@ components:
                 code: PERMISSION_DENIED
                 message: Client does not have sufficient permissions to perform this action.
 
-    Generic404:
+    NotFound404:
       description: Not found
       headers:
         x-correlator:
@@ -571,8 +571,38 @@ components:
                 code: NOT_FOUND
                 message: The specified resource is not found.
 
-    GenericDevice404:
-      $ref: "../common/CAMARA_common.yaml#/components/responses/Generic404"
+    RetrieveQosProfilesNotFound404:
+      description: Not found
+      headers:
+        x-correlator:
+          $ref: "../common/CAMARA_common.yaml#/components/headers/x-correlator"
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 404
+                  code:
+                    enum:
+                      - NOT_FOUND
+                      - IDENTIFIER_NOT_FOUND
+          examples:
+            GENERIC_404_NOT_FOUND:
+              description: Resource is not found
+              value:
+                status: 404
+                code: NOT_FOUND
+                message: The specified resource is not found.
+            GENERIC_404_IDENTIFIER_NOT_FOUND:
+              description: Some identifier cannot be matched to a device
+              value:
+                status: 404
+                code: IDENTIFIER_NOT_FOUND
+                message: Device identifier not found.
 
     Generic422:
       description: Unprocessable Content


RandyLevensalor (approved):
LGTM

eric-murray (approved):

PR #597, opened 2026-08-11
[Changed] Remove device property from SessionInfo and Assignment Info and add to new additional schemas

### LINKED ISSUE #596: Remove `device` property from response body of endpoints that do not accept `device` in request body
**Problem description**
For the quality-on-demand and qos-provisioning APIs, the SessionInfo and AssignmentInfo schema respectively include an optional `device` property intended to be used to disambiguate any multi-property `device` identification included in the request, or otherwise to be provided as a "courtesy" for the API consumer to help them keep track of their sessions / assignments for specific devices.

However the SessionInfo / AssignmentInfo schema are also returned by endpoints that do not accept a `device` property in the request, specifically:
GET /sessions/{sessionId}
POST /sessions/{sessionId}/extend
GET /qos-assignments/{assignmentId}
DELETE /qos-assignments/{assignmentId}

There is never a need to return `device` information for these endpoints, and allowing that opens up the risk that incorrect information may be returned.

In particular, where the APIs created a session / assignment on the basis of IP address(es), these address(es) may have changed by the time one of the above endpoints is called, resulting in a "dilemma" for any API provider who wants to return `device` information to either:
- follow the correct procedure and return the original IP address(es), which have now changed
- or to return the current IP address(es), which is information that the API consumer never provided and maybe should not know.

**Expected behaviour**
Remove this dilemma from the API provider by removing the `device` property completely from the response body of the above endpoints.

Note that this would not be a breaking change for the client, as this property is anyway an optional response property.

**Alternative solution**
Change nothing and rely on API providers to solve the dilemma correctly (by not returning any `device` information for the above endpoints)

**Additional context**
None

eric-murray:
#### What type of PR is this?
* correction

The PR:
- removes the optional `device` property from the `SessionInfo` and `AssignmentInfo` schemas, and creates new schemas `SessionInfoWithDevice` and `AssignmentInfoWithDevice` for use when device disambiguation may be required
- fixes response examples

The `SessionInfo` and `AssignmentInfo` schema are used as the response body for endpoints where there is no requirement to return a `device` identifier.

The `SessionInfoWithDevice` and `AssignmentInfoWithDevice` schema are used as the response body for endpoints where there may be a requirement to return a `device` identifier.

Fixes #596 

The `SessionInfo` and `AssignmentInfo` schemas and can be used by implementations as the session / assignment record if there is no need for the implementation to explicitly record the device identifier.

If the implementation requires to store a device identifier (independent of how the device was identified), then the `SessionInfoWithDevice` and `AssignmentInfoWithDevice` can be used so long as the implementation is disciplined enough not to include `device` in responses where it is never required.

This is not a breaking change for the client, as the PR only removes the possibility to return what is currently an optional parameter for some endpoints.

#### Changelog input
```
 release-note
 - Remove device property from SessionInfo and Assignment Info
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qos-provisioning.yaml
+++ b/code/API_definitions/qos-provisioning.yaml
@@ -396,8 +396,6 @@ components:
       description: Common attributes of the assignment record for a QoS profile assignment
       type: object
       properties:
-        device:
-          $ref: '../common/CAMARA_common.yaml#/components/schemas/Device'
         qosProfile:
           $ref: "#/components/schemas/QosProfileName"
         sink:
--- a/code/API_definitions/qos-provisioning.yaml
+++ b/code/API_definitions/qos-provisioning.yaml
@@ -350,7 +350,7 @@ paths:
                 ASSIGNMENT_UNAVAILABLE:
                   $ref: "#/components/examples/ASSIGNMENT_UNAVAILABLE"
                 ASSIGNMENT_AVAILABLE_WITHOUT_DEVICE:
-                  $ref: "#/components/examples/ASSIGNMENT_AVAILABLE_WITHOUT_DEVICE"
+                  $ref: "#/components/examples/ASSIGNMENT_AVAILABLE_WITH_DEVICE"
         "400":
           $ref: '../common/CAMARA_common.yaml#/components/responses/Generic400'
         "401":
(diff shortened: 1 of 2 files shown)


hdamker:
As written in my [comment](https://github.com/camaraproject/QualityOnDemand/issues/596) on the issue, I wouldn't do the propose changes. In my view they violate the resource model of `session` and several statements we have done across the API and test definition already. See the comment for a proposed simpler solution.

Separately, `SessionInfoWithDevice` is composed from `BaseSessionInfo` rather than `SessionInfo`, so it's missing `sessionId`/`duration`/`status`/etc. — worth fixing in the PR regardless of how the discussion will continue.

eric-murray:
> Separately, SessionInfoWithDevice is composed from BaseSessionInfo rather than SessionInfo, so it's missing sessionId/duration/status/etc. — worth fixing in the PR regardless of how the discussion will continue.

Fixed

PR #602, opened 2026-08-13
Change maximum value for duration to 999 for QoS profiles

### LINKED ISSUE #601: Limit QoS Profiles Duration value to a much lower maximum value
**Problem description**
The current maximum duration value is set to 2147483647.

Current time units defined are Days, Hours, Minutes, Seconds, Milliseconds, Microseconds and Nanoseconds.

As there is no more than a factor 1000 difference between the specified time units, there is no reason for such a high limit. Having such a high limit means there can be many different ways to specify the same number.

Instead, a lower limit would reduce the number of different ways that a given number could be expressed, and would reduce the number of significant digits that could be specified, thus reducing the temptation for API providers to over-specify the precision to which time-based QoS parameters can be measured and controlled.

**Expected behaviour**
Proposal: Limit maximum duration value to 99999, so that between 3 and 5 significant figures can be specified for any time duration between 100 ns and 99999 days, with no change for numbers less than 100 ns.

**Alternative solution**
Don't change the current maximum value

**Additional context**
None

eric-murray:
#### What type of PR is this?
* correction

This PR sets the maximum value of a duration to 999 time units

Fixes #601 

This should not considered to be a breaking change, albeit that the expression of time units for some existing profiles might have to be expressed in a different form (e.g. 1000 Milliseconds becomes 1 Second).

#### Changelog input

```
 release-note
 - Change maximum value for duration to 999 for QoS profiles
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/qos-profiles.yaml
+++ b/code/API_definitions/qos-profiles.yaml
@@ -426,7 +426,7 @@ components:
           example: 12
           format: int32
           minimum: 1
-          maximum: 2147483647
+          maximum: 999
         unit:
           $ref: "#/components/schemas/TimeUnitEnum"
 


hdamker (changes requested):
See my comment on the issue ... I propose to close both the PR and the issue and go with the proposed alternative.

PR #579, opened 2026-06-29
Align quality-on-demand API and tests with Commonalities 0.8.0 (r4.3)

hdamker:
#### What type of PR is this?

enhancement/feature

#### What this PR does / why we need it:

Brings the `quality-on-demand` API specification and its test definitions into line with CAMARA Commonalities 0.8.0 (r4.3). The change is backward-compatible with the released `1.1.0`: each tightened constraint makes an already-intended, domain-limited bound machine-readable and rejects no previously-valid request.

Highlights:
- Reuse the synced common components via `$ref` into `code/common/` at each use-site: `openId`, `x-correlator`, `Device` / `DeviceResponse`, `ErrorInfo`, and the generic error responses. API-specific code-subset error responses are kept local.
- Embed the mandatory `info.description` templates (authorization-and-authentication, additional-error-responses, request-body-strictness) and set `x-camara-commonalities: 0.8.0`.
- Add OWASP-aligned constraints (`maxLength` / `maxItems` / `minimum` / `maximum` / `format` / `pattern`) to the remaining local schemas.
- Align the QoS status-change notifications to the CloudEvents model: consume the shared `CloudEvent` and `SinkCredential` from `CAMARA_event_common.yaml`, express the event as a CloudEvent subtype with its own event-type enum and discriminator, and adopt `ACCESSTOKEN` + `PRIVATE_KEY_JWT` (adding `PRIVATE_KEY_JWT_NOT_CONFIGURED` to the create-session `422`).
- Align the five `quality-on-demand-*.feature` test definitions (repoint the renamed device schema, add `invalid_x-correlator` and `429` scenarios, quote scope values).

The event and notification changes (the single semantic change) are isolated in their own commit, separate from the metadata/reuse/constraints/tests, so each part can be reviewed independently.

#### Which issue(s) this PR fixes:

Fixes #571

#### Special notes for reviewers:

- Backward compatibility: every change is backward-compatible with the released `1.1.0`, so each addition is constrained by Commonalities #658. Dropping `PLAIN` / `REFRESHTOKEN` from the `SinkCredential` enum is a non-breaking restatement (both were already documented as disallowed); `PRIVATE_KEY_JWT` is additive.
- `duration` intentionally has no API-level maximum (the limit is the QoS Profile `maxDuration`); `maximum: 2147483647` is the int32 ceiling, added only to satisfy the linter without restricting any valid value.
[...]

PR #619, opened 2026-09-15
Align quality-on-demand and qos-provisioning with Commonalities r4.4

### LINKED ISSUE #585: quality-on-demand.yaml: replace local SingleIpv4Addr/SingleIpv6Addr with common $ref (after Commonalities r4.4)
#### What type of issue is this?

* enhancement/feature

#### Problem description

`quality-on-demand.yaml`'s `ApplicationServerIpAddressListType.items.anyOf` uses two locally-defined schemas that duplicate `code/common/CAMARA_common.yaml`:

- `SingleIpv4Addr` is already byte-for-byte identical to the common `SingleIpv4Address` schema (same description, `maxLength: 15`, `format: ipv4`) — just kept under the old pre-cleanup name.
- `SingleIpv6Addr` will become identical to the common `SingleIpv6Address` schema once [Commonalities#654](https://github.com/camaraproject/Commonalities/pull/654) ("Define SingleIpv6Address schema for general use") merges (same description, `maxLength: 45`, `format: ipv6`).

#### Possible evolution

Once Commonalities r4.4 is published (expected to carry #654), replace both local schemas with `$ref`s into `../common/CAMARA_common.yaml#/components/schemas/SingleIpv4Address` and `.../SingleIpv6Address`, and delete the two local definitions. Non-breaking — the schema shape is identical, only the reference changes.

#### Alternative solution

Fix `SingleIpv4Addr` now (the common equivalent already exists) and handle `SingleIpv6Addr` in a second pass once r4.4 ships. Not preferred: two touch-ups instead of one, and the file would briefly mix `$ref` and local styles for the same concept.

### LINKED ISSUE #618: Align quality-on-demand and qos-provisioning with Commonalities r4.4
**Problem description**

`release-plan.yaml` declares `commonalities_release: r4.4` for release r4.2, and the r4.4 common files are now in `code/common/` (#616), but both API definitions still declare `x-camara-commonalities: 0.8.0` and still follow r4.3 patterns. This issue collects the points that need to be addressed to bring `quality-on-demand` and `qos-provisioning` to Commonalities 0.9.0 (r4.4).

Note that CI validation is currently green except for two `P-027` warnings. Most of the items below are not covered by any validation rule — they come from the r4.4 changelog and Design Guide — so a passing pipeline should not be read as r4.4 alignment.

Scope is the two API definitions in this repository. `qos-profiles` is aligned separately in [QoSProfiles](https://github.com/camaraproject/QoSProfiles).

**Expected behavior**

All changes below are backward compatible. Where r4.4 offers a narrower replacement than what an operation declares today, the declared error code set is preserved rather than narrowed.

*Metadata and documentation*

- [ ] Set `x-camara-commonalities: 0.9.0` in both specs (r4.4 `VERSION.yaml` is `0.9.0`; Design Guide §5.3.7).
- [ ] Re-copy the `additional-error-responses` mandatory `info.description` block from `code/common/info-description-templates.yaml` into both specs. This is the `P-027` warning. Paragraph 3 changed in Commonalities#693: the sentence "The applicable Commonalities Release can be identified in the `API Readiness Checklist` document associated to this API version." becomes "The applicable Commonalities Release can be identified from the `x-camara-commonalities` field, the changelog and the metadata of the released API version."
- [ ] Bump the r4.3 Design Guide links in both specs to r4.4.

*Error responses — migrate off the deprecated `Generic<status>` responses*

r4.4 deprecates all `Generic<status>` responses in favour of a minimal named catalogue plus a shared example pool, and plans their removal next cycle (Commonalities#665).

- [ ] Replace with the catalogue entry where the code set is identical: `Generic400` → `BadRequestWithRange400`, `Generic401` → `Unauthenticated401`, `Generic422` → `DeviceIdentifier422`, and `Generic410` in the notification callback → `CAMARA_event_common.yaml#/components/responses/SinkGone410`.
- [ ] Keep a local 429 declaring both `TOO_MANY_REQUESTS` and `QUOTA_EXCEEDED`, referencing the common examples. `TooManyRequests429` declares only `TOO_MANY_REQUESTS`, so referencing it directly would drop `QUOTA_EXCEEDED` from the documented contract.
- [ ] Keep a local 404 declaring both `NOT_FOUND` and `IDENTIFIER_NOT_FOUND` on the device-identified operations, referencing the common examples. r4.4 splits `Generic404` into `NotFound404` and `IdentifierNotFound404`. Whether `NOT_FOUND` can be dropped on `retrieveSessionsByDevice` is the subject of #592.
- [ ] Replace the local `PermissionDenied403` and `NotFound404` definitions in both specs with references to the common responses of the same name — r4.4 adopted the same names and the same code sets.
- [ ] In `quality-on-demand.yaml`, replace `GetSessionBadRequest400` and `DeleteSessionBadRequest400` with the common `BadRequest400`, and `GenericExtendSessionDuration400` with the common `BadRequestWithRange400`.

*Error examples — use the new shared example pools*

r4.4 adds a `components/examples` section to both common files. Every error response in both specs currently carries inline examples.

- [ ] Reference the common examples from the responses that stay local — `CreateSessionBadRequest400` / `CreateAssignment400` and `CreateSessionUnprocessableEntity422` / `CreateAssignment422`. The `INVALID_CREDENTIAL`, `INVALID_TOKEN`, `INVALID_SINK` and `PRIVATE_KEY_JWT_NOT_CONFIGURED` examples live in `CAMARA_event_common.yaml`; the rest in `CAMARA_common.yaml`.
- [ ] Keep inline only what has no common example — the 409 `CONFLICT` responses (see below) and API-specific codes.

*Common schema reuse*

- [ ] Replace the local `SingleIpv4Addr` and `SingleIpv6Addr` in `quality-on-demand.yaml` with references to the common `SingleIpv4Address` and `SingleIpv6Address`, and delete the local definitions. This resolves #585 — its precondition is met, since Commonalities#654 landed in r4.4. Constraints are identical; only the IPv4 example value differs.
- [ ] Consider referencing the new common `CAMARA_event_common.yaml#/components/schemas/Sink`, which is constraint-identical to the local `sink` string in both specs.

*Test definitions*

- [ ] Normalise the OAS pointers in the `.feature` files to the `#/` prefix used by the r4.4 test template (Commonalities#678, #683). The files currently mix `#/components/schemas/X` and `/components/schemas/X`, sometimes within the same file.

**Alternative solution**

Two r4.4 changes are deliberately **not** proposed here, because both are breaking and would require a new major version rather than the planned `quality-on-demand` 1.2.0 and `qos-provisioning` 0.4.0:

- `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED`. The r4.4 Event Subscription and Notification Guide requires an implicit-subscription API to reject a `sink` it cannot honour with this code instead of silently ignoring it. Both APIs offer implicit subscriptions with an optional `sink`. The guide states that adding the code to an API that already offers implicit subscriptions is a breaking change requiring a new major version, and leaves the timing to the API. Suggest a separate enhancement issue per API.
- `409 CONFLICT` → `INCOMPATIBLE_STATE`. `CONFLICT` is deprecated in the Design Guide, and r4.4 removes it from the common `Generic409` enum and deletes its example. `SessionInConflict409` and `AssignmentConflict409` still use it, so they keep a local definition with an inline example for now. Changing the code string is breaking.

**Additional context**

One further point needs a Commonalities decision rather than a change here. The new (informative) Appendix B restricts `operationId` verbs to an approved list, and `extendQosSessionDuration` and `revokeQosAssignment` use verbs that are not on it. Appendix B prescribes the remedy itself — submit the verb for formal review via a Commonalities issue — and renaming an `operationId` would break generated clients, so the suggestion is to raise that issue rather than rename. Callback `operationId`s are explicitly out of Appendix B scope, and every other Appendix B rule already holds for both specs.

Verified as requiring no change: the new 53-bit numeric bound (the largest value in either spec is `2147483647`); the `Config` → `ConfigBase` refactor and `CreateSubscriptionDetail` removal (neither spec references either); the common `Pagination` schema (neither spec paginates); the ICM documentation link bump to r4.2 (no r3.3 links present); `x-correlator` moving from SHOULD to MUST (both already support it).

Predecessor for r4.3: #571.

A PR implementing this will follow.

hdamker:
#### What type of PR is this?

correction

Brings `quality-on-demand` and `qos-provisioning` into line with CAMARA Commonalities 0.9.0 (r4.4). Backward-compatible with the released `1.1.0` / `0.3.0`: no response code set is narrowed and no existing error `code` value is renamed.

Highlights:
- Set `x-camara-commonalities: 0.9.0` in both specs and re-copy the drifted `additional-error-responses` `info.description` block.
- Migrate off the deprecated `Generic<status>` responses onto the r4.4 catalogue: `$ref` the catalogue entry directly where its code set matches exactly (`BadRequest400`, `BadRequestWithRange400`, `Unauthenticated401`, `PermissionDenied403`, `NotFound404`, `DeviceIdentifier422`, `SinkGone410`); where it would narrow the declared set, keep a local response with the same codes and `$ref` the matching shared examples instead (a combined 404 for the device-identified operations, a combined 429 with `QUOTA_EXCEEDED`).
- Source error examples from the new `CAMARA_common.yaml` / `CAMARA_event_common.yaml` example pools wherever a local response stays.
- Reuse the new common `SingleIpv4Address` / `SingleIpv6Address` schemas in place of local duplicates (resolves #585), and the new common `Sink` schema in place of the local inline definition.
- Normalize the `.feature` test-definition OAS pointers to the `#/` prefix used by the r4.4 test template.

Not carried in this PR, because both would require a new major version: adding `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` for the optional `sink` on session/assignment creation, and replacing the deprecated `409 CONFLICT` with `INCOMPATIBLE_STATE`.

Fixes #618

- Backward compatibility: every declared response code set is unchanged; the two responses that would otherwise narrow (`404`, `429`) stay local with their full existing code set rather than adopting the narrower catalogue entry.
- Two example texts changed by adopting the common ones: `INVALID_CREDENTIAL`'s message now says "Only Access token or Private key JWT are supported" (was stale from before #579 added `PRIVATE_KEY_JWT`), and `INVALID_TOKEN`'s message now describes an invalid/expired token rather than "Only bearer token is supported". Only the `code` is contractual; the message text is not.
- `qos-provisioning`'s local `sink` schema's description already matched the new common `Sink` schema's description verbatim; `quality-on-demand`'s kept its more specific session-termination wording via the sibling `description` next to the `allOf`.

#### Changelog input

```
 release-note
Align quality-on-demand and qos-provisioning with CAMARA Commonalities 0.9.0 (r4.4): migrate off the deprecated Generic<status> error responses onto the new catalogue and shared examples, reuse the new common SingleIpv4Address/SingleIpv6Address and Sink schemas, and fix the additional-error-responses template drift. Backward-compatible.
```

#### Additional documentation

This section can be blank.

```
docs

```


hdamker:
@camaraproject/quality-on-demand_codeowners please let me know if there are any changes needed to this PR

